PR #10, opened 2023-12-07
Feat: Fix Chat Signing

ghost:


Dueris:
I'm confused. Whats broken?

ghost:
You get kicked out cause of chat signing

Dueris:
What?

ghost:
like this 
image

ghost:
very bad image but says stuff related to chat signature

Dueris:
That's due to a weird error that happens when the client breaks itself. If it's not verified it will still break things. Just rejoin the server to fix it.

ghost:
ik but its annoying to rejoin

Dueris:
Would it be more annoying to have every chat message possibly break or to spend 5 seconds to rejoin

ghost:
well if it breaks that cause issues so ig rejoin

Dueris:
Then is there a need for this patch if it's just going to further break things and cause more of a hassle?

ghost:
no

PR #18, opened 2024-04-07
ThreadRipper

Dueris:
Adds lots of optimizations related to Threading that require lots of threads to make Canvas basically fly.


Dueris:
Being pushed with more upstream directly to 1.20.4, no need for this

PR #23, opened 2024-06-14
Fix funding.yml --skip

ItzPancakse:


ItzPancakse:
not spose to have link only the part after _`https://ko-fi.com/`_

PR #59, opened 2025-01-06
Fixed building

MrMasrozYTLIVE:
This PR fixes GH Actions building for:
1) Original repo (By downgrading gradle back to 8.7 from 8.12)
2) Any forks (By adding an IF check for `send discord notifications` task)


MrMasrozYTLIVE:
Uhhhh might want to check for publishing if token is set haha

PR #62, opened 2025-01-10
Fixed some bugs

MrMasrozYTLIVE:
Changes:
Fixed ref list crash by checking if endIndex < 0
Fixed launching server through :canvas-server:runDevServer task (Launcher now uses system java instead of JetBrains JDK 21) Added debug logging for broken packets

### DIFF
--- a/patches/server/0004-Threaded-Dimensions.patch
+++ b/patches/server/0004-Threaded-Dimensions.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Threaded Dimensions
 
 
 diff --git a/build.gradle.kts b/build.gradle.kts
-index ebbf963582bf89e13a3048f0838038debd1ab756..0af5fe5c1a038c7675df9b26449a81a907f37f91 100644
+index bf1eccdc5b4553f2b25a3049c468a35e64cce3dc..795ff47f075eff2699819a0cc6bac799682b0c0e 100644
 --- a/build.gradle.kts
 +++ b/build.gradle.kts
 @@ -43,7 +43,7 @@ dependencies {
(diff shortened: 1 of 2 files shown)


patches/server/0004-Threaded-Dimensions.patch
```diff
@@ -1530,6 +1536,19 @@ index 1a37654aff9a9c86c9f7af10a1cf721371f0c5ec..4d5aff469934bd2b401b68d4d2db8a04
...
+index 6ba6933bec2a4d82df29fa2f571f0ef76902ef56..5f395f5b1af5b6018b1cfd3b317ae538a90f3670 100644
+--- a/src/main/java/net/minecraft/network/protocol/game/ServerPacketListener.java
++++ b/src/main/java/net/minecraft/network/protocol/game/ServerPacketListener.java
+@@ -11,6 +11,7 @@ public interface ServerPacketListener extends ServerboundPacketListener {
+ 
+     @Override
+     default void onPacketError(Packet packet, Exception exception) throws ReportedException {
+-        LOGGER.error("Failed to handle packet {}, suppressing error", packet, exception);
++        LOGGER.error("Failed to handle packet {} on thread {}, suppressing error", packet, Thread.currentThread().getName(), exception);
++        new Throwable().printStackTrace();
```
Dueris, line 1549: why make a new throwable? there is an `Exception` argument already provided

Dueris (changes requested):

patches/server/0004-Threaded-Dimensions.patch
MrMasrozYTLIVE: Because it has no stacktrace

MrMasrozYTLIVE: That exception is instance of net.minecraft.server.RunningOnDifferentThreadException which overrides stacktrace to be empty

Dueris: ahh, gotcha.

Dueris (approved):

PR #75, opened 2025-02-10
docs: more readme improvements

ghost:
Added:
- Requirements for running CanvasMC with explanation why Java 22 is minimum
- `runDevServer` command to scripts
- Rephrase support section a little bit

Before merging I'd like to know if I'm missing any more important scripts and if I got FFM explanation right

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,10 +21,10 @@ Useful links:
 
 ## Requirements
 
-### Running Canvas
-
 > [!WARNING]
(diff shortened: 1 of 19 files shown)


Dueris:
`./gradlew rebuildPatches` no longer exists. Thats why the `./rebuildPatches` script exists, since rebuilding has 2 file formats and 3 directories, so 12 tasks total for rebuilding. the script i made scans the directories for modifications and runs fixup and rebuild on modified dirs. when the script is appended with `--force`, then it runs all directories

PR #81, opened 2025-03-08
Update egg-canvas.json

kerpsonn:


Dueris:
After some discussion, we have decided we are going to move the Canvas egg to a gist and add it to a new page in the [docs](https://docs.canvasmc.io/). We will merge your changes with the new gist/doc, thank you for your contribution :)

PR #109, opened 2025-07-03
Update README.md to reflect current status of Canvas

ccorp2002:
Hi, dueris. Let me know if anything needs changing, or feel free to edit this yourself.


Dueris:
useful links doesnt need to be marked as "old", they will still be used

also can we add "remove folia region profiler" to the TODO?

Dueris:
thank you for the contribution

PR #111, opened 2025-07-10
Build adjustments

Toffikk:


Dueris:
Ahh thanks for catching these lmao. I made the spark support patches on my Folia local branch so I forgot to rename them to the right patch numbers.

PR #112, opened 2025-07-12
Apply ontop of Purpur

Toffikk:
This will require extensive testing to make sure everything works as it should. 
Right now the obviously broken thing that we have to figure out how to fix is max joins per second from purpur, other than that everything looks fine(?).

Testers would be greatly appreciated!

TODO:
Also fix tpsbar/rambar probably forgot to change it (crashes because of that on startup)


Dueris:
ignore my requests, reasoning within development channel of the discord server

Dueris:
what may end up happening is TPS bar will be in Canvas itself, as I do want to make major adjustments to it..

Dueris:
RAM bar is fine, idgaf, but TPS bar I want in Canvas natively

Dueris:
Are we still doing anything with this PR?

Toffikk:
I think this should be put on pause rn to be picked up later, to decide what we pull/what we don't pull

PR #117, opened 2025-07-27
Optimize Packet Processing

Dueris:
This optimizes packet processing through using the intermediate tasks system in the tick region scheduler, allowing us to process player packets in-between ticks, rather than scheduling to the entity scheduler and running those when the entity scheduler is ticked. A similar system was made in SpottedLeafs `beta/new-scheduler` branch in Folia, however this was flawed and bugged in many ways. This fixes those bugs and attempts to improve upon that system, given the new scheduler heavily impacted performance in very good ways.

This needs to be heavily tested before merged into the `ver/1.21.8` branch.


Dueris:
We tested for the specific bug we were looking for, this is no longer replicatable, this will be merged to `ver/1.21.8` soon

PR #123, opened 2025-08-15
Secure seed ported from Leaf

omdumrotat:
This PR adds the secure seed function that is ported from Leaf (https://github.com/Winds-Studio/Leaf/blob/ver/1.21.8/leaf-server/minecraft-patches/features/0132-Matter-Secure-Seed.patch, https://github.com/Winds-Studio/Leaf/blob/ver/1.21.8/leaf-server/minecraft-patches/features/0133-Matter-Secure-Seed-command.patch). It is useful for server owners who doesn't want to get their seed exposed via seed cracking.


Dueris:
Im not entirely comfortable adding secure seed into Canvas currently, as from what Ive seen is that this option breaks world gen datapacks, and for people moving from Folia -> Canvas, those datapacks are often times the only way to add those worlds. By implementing secure seed, we risk people enabling this thinking "oh look fancy security patch" when it will just break their worlds. If this were to become stable && not break world gen datapacks, I would be more inclined to adding this. However for the time being I do not wish to implement this.

PR #131, opened 2025-09-22
Region Profiler V2

Dueris:
This PR introduces Region Profiler V2, a rewrite of SpottedLeaf’s original profiler, adding full Spark integration for Folia’s region threading environment.

---

# Region Profiler V2

In the Folia region threading environment, there are multiple issues with utilizing this environment. One of the core issues is that profiling regions independently is nearly impossible unless you have a single region loaded on a server. For production environments, this is oftentimes not possible.

To combat this, **SpottedLeaf**, the creator of V1 of the region profiler, implemented a more generalized profiler to determine points of lag. The issue with this is that it had no detailing whatsoever, causing issues when trying to make genuine performance optimizations for the dedicated server, or for server owners and developers to find sources of lag.

**Spark** is the builtin profiler for Paper, upstream of Folia, and gives a perfect amount of data to help diagnose issues and performance problems. However, the way Spark works prevents us from being able to track regions independently. Spark can track specific **threads**, not **regions**, as regions are tasked based on a scheduled task thread pool.

To combat this, we created a **task pinning** system that allows runners to become marked as dedicated processors for a selected area of chunks to be profiled.

---

## Threading Details

As stated in the brief explanation prior, Spark can track specific **threads**, not **regions**, and so we need a way to ensure that the region is on a specific thread at all times. An API was introduced to allow **pinning** the area of chunks to a dedicated tick runner.

This does include some performance penalties, as since the area needs isolation, it cannot be utilized in scaling the server, so other tasks cannot use the thread for task processing while a region is pinned to it. For the spark report to be accurate, we need to make the area completely isolated. The new API allows us to mark tasks as pinned to the thread id of a tick runner, which marks the thread as only able to process tasks pinned to it. The API internally allows us to pin any amount of tasks to any amount of threads, however we limit the server’s usage to only a single area at a time, as Spark can only run one profiler at once.

This should never be run 24/7 during production, this is a tool purely for diagnostic purposes, as this does limit scaling a bit.

When the relationship between the thread runner and the tick task is formed or destroyed, the scheduler will view this as a filter to ensure that the tick task is only ticked on the thread runner, even if it exceeds the steal threshold. This is ensured by applying a few behavioral changes in the forms of filters:

* If a pinned task is on a different thread than the one it's being pinned to, it will not be picked up by anything until it has exceeded its steal threshold, where the pinned thread will take the task.
* If the tick thread runner gets assigned a pinned task, it will only run the pinned tasks, and any other tasks on the current thread will eventually exceed their steal threshold and be picked up by other runners.

---

## Region Logic

Regions behave unpredictably due to player interaction with the world. Any server could do anything to the region being profiled. It may be loaded already, it may not be loaded at all. So when pinning, we load the chunks in the area via the ticket
`TicketType#REGION_PROFILING_HOLD (canvas:region_profiler_hold)`, which acts as a non-persistent ticket that is kept throughout runtime or until told to be removed, like the `forced` ticket. Once all the chunks are loaded, we pin the running region to a thread (any available).

### Splits

Splits introduce complexity. Given chunks can be loaded and unloaded, the area being profiled is kept loaded. This means we can encounter splits, however they can also happen for a wide number of other reasons. So, we store the chunks being profiled in `SparkRegionProfilerExtension#PROFILING_CHUNKS`, so when a split occurs we can mark the region with those chunks as the one needing to be pinned. The original one will be unpinned, and profiling will continue as normal.

### Merges

When a region is merged, we unpin the region and then repin the resulting region.

### Region Death

Upon death of a region, we immediately fail-fast and kill the server. This should **NEVER** happen under any circumstances.

The fail-fast model is intentionally aggressive: developers and server owners are guaranteed that profiling results are either valid or the server halts, with no undefined state in between. This allows us to catch bugs much more quickly.

---

## Spark Integration

We modify the class `me.lucko.spark.paper.common.command.modules.SamplerModule` in Spark to integrate a new argument, `--region` to the spark profiler command. This extension takes 2 or 4 arguments, the `from` and `to` block positions of the region.

* Start: `/spark profiler start --region ~ ~`
* Stop: `/spark profiler stop`

The `--region` extension prepares the server for isolated profiling by defining an area to profile and pinning its region. This does **not** replace or disable the existing `spark` command, but is designed to work **with** it to support targeted profiling.

When executed, the extension first checks that the requested profiling area lies within the world border and that the number of chunks does not exceed `512L` (a limit which may change in the future). It then loads the chunks asynchronously if not loaded already using
`ServerLevel#moonrise$loadChunksAsync(int, int, int, int, Priority, Consumer)` at `Priority#HIGHER`. Once loading completes, the callback already ensures all chunks are fully available in a region, so we place profiler tickets, and pin the region for profiling.

Utilizing our spark plugin internals, we create an interchangeable system that swaps between REGEX pattern-based matching and thread name matching, since `SparkRegionProfilerExtension#TRACKING_THREAD` can provide us with the `ScheduledTaskThreadPool.TickThreadRunner` we are actively profiling. When there are no regions pinned, the system defaults to the standard REGEX pattern:

```
"Region Scheduler Thread #\d+"
```

---

## Concurrency & Safety

* Use copy-on-write structures to avoid mutation during active reads.
* All profiler state is bound to volatile references, ensuring visibility across threads without explicit synchronization.
* Non-persistent tickets are attached at the chunk level, preventing accidental unloading of a profiled region.
* Hard server shutdown on an unknown issue prevents corruption of profiling output.

---

**Author**: Dueris
**Version**: 2.0

---

Example Usage:
`/spark profiler start --region ~ ~ -47493 53144 --timeout 300` - https://spark.lucko.me/s0jWaRaZSK
Do note, this example usage is to just show an example output, this is not a genuine performance benchmark.

---

This PR includes numerous cleanups too to the Canvas patches, including condensing the "Rewrite Folia Scheduler" patch and the "Rewrite Moonrise Executor" patch. This also fixes numerous very important issues.


Toffikk (approved):
lgtm

TheDeafCreeper:
Looks pretty cool and will definitely be useful. Since there's a performance hit it might be a good idea to have some sort of message for opped users (and have a permission node) when they join while a profiler is running for a specific region, this I can't imagine this is a particularly big issue though.

R00tB33rMan (approved):
Looks good to me — nice work with the cleanup. Keeps things concise when review is needed in the future

yunuservices:
This pr will be too much useful for server owners. I really liked the idea and usage of it. Since the old system was showing you general instead of regionized.

Toffikk (approved):

MachineBreaker:
> This pr will be too much useful for server owners. I really liked the idea and usage of it. Since the old system was showing you general instead of regionized.

--not-combined on spark shows you individual regions

Dueris:
MachineBreaker, no it does not. That argument ungroups the *threads*. Spark has never been able to do region profiling, ever. Its profiler has only ever been able to track threads, not regions.

Dueris:
MiTask marked as approved in our DMs. This will be merged shortly

PR #134, opened 2025-10-19
Rewrite Waypoints

Dueris:
Replaces the ServerWaypointManager implementation with RegionThreadingWaypointManager. This implementation essentially is a region-threading-safe version of the ServerWaypointManager, with a few specifications:
- All methods must be called on specific thread contexts, specifically Players and such.
- Probabilistic distance throttling for waypoint updates to prevent spam of entity tasks and updates when players are spread out
- Removes the usage of the Table in the original waypoint manager, given this is slow af
- Generally optimize things a TON with this, however should still be optimized further for the sake of trying to make this global state as fast as possible

This implementation also fixes a few issues with the old region-threading waypoints system, and improves upon it being fully Vanilla compatible.


canvas-server/minecraft-patches/sources/net/minecraft/server/waypoints/ServerWaypointManager.java.patch
```diff
@@ -1,37 +1,48 @@
 --- a/net/minecraft/server/waypoints/ServerWaypointManager.java
 +++ b/net/minecraft/server/waypoints/ServerWaypointManager.java
-@@ -22,11 +_,17 @@
+@@ -22,11 +_,20 @@
  
      @Override
      public void trackWaypoint(WaypointTransmitter waypoint) {
 -        // Folia - region threading
-+        if (io.canvasmc.canvas.Config.INSTANCE.disableWaypointsFix) return; // Canvas - region threading - restore waypoints
++        if (true) throw new UnsupportedOperationException("Unsupported in region threading");
```
MrMasrozYTLIVE: Is this if(true) needed? (Not like I have anything against it, just don't understand what's the point of it if everything else after the if statement is commented anyway)

MrMasrozYTLIVE (approved):
Lgtm

canvas-server/minecraft-patches/sources/net/minecraft/server/waypoints/ServerWaypointManager.java.patch
Dueris: Honestly I did that because the commenting out didn't happen until after that, and I wanted to keep the original code for just seeing the Vanilla code during development for reference

MrMasrozYTLIVE: Reasonable

MrMasrozYTLIVE (approved):

R00tB33rMan (approved):
LGTM with applied changes

yunuservices:
looks good. 😶

Toffikk (approved):

Dueris:
The [docs page](https://docs.canvasmc.io/threadedregions/fixes#waypoints--the-locator-bar) needs to be updated to reflect these changes

PR #137, opened 2025-10-29
Entity Culling

Dueris:
This implementation tracks and untracks entities out of sight of the Player entity via raytrace culling. This can greatly improve network bandwidth, and even client FPS.

This implementation creates a system where if entities cannot be seen by a player, specifically via a non-translucent block between teh entity and player for example, it gets removed from tracking of that player.

Note: the entity visibility is conducted via an 8-point raytrace calculation using the entity AABB, slightly inflated so that it covers edges of walls better, to make the effect more smooth.

The system *does* come with pre-filters so that Vanilla compat is kept. Those filters are: (for this example we will have "Target" be the entity that is being tested for culling, and "Origin" as the player we are testing to)
- If the origin is *not* a real player(decided by ServerPlayer#isRealPlayer), we cull them.
- If the origin is a spectator, we don't cull any entities, as trying to cull for spectator mode sounds like a personal nightmare and a half
- If the target is a Player, if their gamemode is spectator we will cull them. Do note, we have already ensured by this check we are not in spectator mode either, so they wouldn't be shown regardless.
- If the target is a player though and has passed the previous checks, they will not be culled.
- If the target is glowing, we do not cull them. The glowing effect can be seen through walls.
- If the target is an armor stand, and is invisible, it may be used as a "marker entity", so we don't cull them.
- If the target has a custom name(nametag) always visible(possibly via a plugin), we shouldn't cull them, as nametags can be displayed through walls.
- If the target is a display entity, we do not cull it.
- If the target is within 3 blocks, we don't cull them
- If the target is >=64 blocks out, we cull them
- If the target bounding box(inflated by 0.3) is greater than 20 on all 3 axis's, we will not cull them
- If the origin can see the full target AABB using the eye position as its look origin, we do not cull them. Otherwise, they are culled.

> [!NOTE]
> This should probably be tested further for performance regressions, visual appeal, actual effect networking-wise, etc. Generally just needs more testing before I feel comfortable with merging this to the master branch.

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/server/level/ChunkMap.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/server/level/ChunkMap.java.patch
@@ -25,6 +25,15 @@
                  this.ticketStorage.getTicketDebugString(longKey, true),
                  this.distanceManager.getChunkLevel(longKey, true),
                  optional1.<Integer>map(chunk -> chunk.getBlockTicks().count()).orElse(0),
+@@ -1277,7 +_,7 @@
+                 double d1 = vec3_dx * vec3_dx + vec3_dz * vec3_dz; // Paper
+                 double d2 = d * d;
+                 // Paper start - Configurable entity tracking range by Y
+-                boolean flag = d1 <= d2;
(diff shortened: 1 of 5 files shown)


Dueris:
This is being closed. Generally, this works fine. But with high-ping players this becomes unsafe and unideal.

Issue #133, opened 2025-10-15
Feature request: Sentry support


NicklasMatzulla:
Hello,

I would like to request that you integrate support for [Sentry](https://sentry.io). Sentry is a platform designed to simplify error logging. I personally switched from [Luminol](https://github.com/LuminolMC/Luminol) to Canvas because Canvas meets all of Folia's requirements, unlike other forks, which can lead to critical errors/led to critical errors for me. Apart from that, support for Sentry would also be very beneficial for Canvas, and I would be delighted if this could be integrated in the near future. In theory, there are already some forks that support Sentry, such as Luminol mentioned above, which has Sentry support from Leaf (if I'm not mistaken).

Thank you for your help.

Best regards

Toffikk:
Generally i would be in favour of that, but we’d have to implement sentry in a better way than it is done in pufferfish (and by extension leaf probably) so it contains more useful information and a more recent version of it, not 5.x. We could take a look at the open pufferfish pr for one of those things like the additional info, although it still might be lacking in my opinion

Toffikk:
This will be looked into by us soon

PR #139, opened 2025-11-14
Added sentry logging implementation

NicklasMatzulla:
This PR adds support for Sentry logging, as requested in feature request https://github.com/CraftCanvasMC/Canvas/issues/133.


Dueris:
Requested review, will look at this closely when I have time. Thank you for the PR

Toffikk:
Generally, while this does resolve that issue, we came to a conclusion that we don’t see the benefit in such a “bare” implementation, but instead prefer to have a more complete one with better debugging implemented, which I planned to implement myself but haven’t had the time to do that yet. And as this is not an urgent issue i’m not in favour of pulling this atm, if you want to however put work into this to implement more features, feel free to reach out in our discord to discuss things in more detail.

Dueris:
Yea Toffikk is right, I honestly forgot about that conversation lol. But yea, having further debugging would be handy, otherwise I am not entirely fond of pulling this.

R00tB33rMan (changes requested):
As a not too essential side note, it’d be best to bump to SnakeYAML 2.5 if you’re looking to bump it in the first place

NicklasMatzulla:
I created this PR only as a quick fix to solve the current problem  that errors are not being logged in Sentry. I would recommend merging the PR and improving the implementation at a later date, if time allows. I am aware that my implementation ignores many features, but my priority was to have a solution in place. After all, it's better to have something than nothing, especially if you don't have time.

Toffikk:
Sentry support however is not something that is urgent, especially given that all of the info this impl provides is also what we can from logs more easily with even more details, and sentry not being implemented isn’t something we can qualify as a problem, only as a feature request.

Toffikk:
Also i don’t think proper credits have been given here as the patch and some things seem to be taken from Luminol(?) but only mention pufferfish.

Toffikk (changes requested):
Please update the library to 8.26.0 (latest atm) and provide proper credits for the implementations

coredex-source (commented):
I personally don't think that this PR should be merged without a better implementation and as Toffikk pointed out it seems to not have proper credits. I don't think sentry is "required" and I'd personally like to see a feature rich implementation instead.

Dueris (changes requested):
I would like a more feature rich implementation of this, and then we can work on merging the PR. It isnt urgent, and we would like generally more debugging, as this doesnt rly tell us much, and we can get more info from logs and crash reports and spark.

Also, I would like to mention that just throwing an exception in the test plugin isnt helpful, I understand why you did it, but I personally think we shouldn't intentionally kill the plugin bc then we lose the rest of its tests. It can be removed later sure, but I just dont see a point as it just seems inconvenient

Toffikk:
For the record, a feature that would make this PR more viable would be figuring out how to attach configuration files to the sentry environment report

Toffikk:
@NicklasMatzulla Do you plan on updating this pr in accordance with the comments raised here?

PR #140, opened 2025-11-16
annotate the api module

Toffikk:
### Motivation
These changes had to be made in order to fix the failing tests, everything is now assumed not null by default unless specified otherwise, and also the jankson classes were a hot mess as half the code wasnt annotated or was wrongly so

### TODO BEFORE MERGE:
 - [x] ensure that API has been annotated correctly, cc @Dueris
 - [x] check if all annotations are placed correctly in the *jank*son classes
 - [x] test if i didnt change some behaviour by accident
   - shouldnt be the case
   
### Notes
Some parts of the code got cleaned up in the process but have retained their original behaviour, null annotations have been placed in the respective `package-info.java`'s


Toffikk:
@Dueris all annotations should be placed correctly now, i just need a glance at the api and then you can approve and merge this

Dueris:
Im guessing most of these changes are also cleanup?

Dueris:
Asking because during my review I noticed a lot of changes non-annotation related.

Are you sure each cleanup change u made is correct?

Toffikk:
yeah, should be, its just removing redundant null checks handled by instanceof checks and switching to pattern matching and switch cases

Dueris (approved):

PR #143, opened 2025-11-22
Rewrite Folia Scheduler

Dueris:
Canvas has been running the SpottedLeaf beta scheduler for Folia since V3 was created. This implementation caused numerous issues, which have been mostly fixed, however there is still the simple fact that these issues existed, concerning.

This rewrite is intended to base off the original Folia scheduler, and implement what the beta scheduler was set out to do. Those things being:
- Thread hopping being absurd
- Mid tick tasks never being called

Essentially, the beta scheduler was fundamentally flawed at its core. It's entire design is flawed, which causes many issues, like https://github.com/PaperMC/Folia/issues/346, and https://github.com/PaperMC/Folia/issues/341, and much, much more. The old scheduler on the other hand, by ConcurrentUtil, has a more clear structure and lifecycle, and actually doesn't break everything...

This implementation of the old scheduler, takes the scheduler and adds back mid-tick-tasks as a form of "ticking" the task, as it does force the region to act as if it's being ticked, and that it is in the state of ticking, instead of parking until the deadline for that tick. It also includes a revised implementation of task pinning, and functions much better than before. Thread hopping for tasks has also been improved on, by sectioning off the queue into a "global" queue, and integer-keyed partitions, allowing assigning tasks to a thread id to try and limit what threads can see a task, making it more consistent on what thread the task is run on. This also does include making tasks "globally visible", even when marked in a partition of the queue, which only happens if the task deadline is missed by 2ms. The "global" queue anyone can take from, same as "globally visible" tasks. This helps ensure thread hopping is limited, and the 2ms deadline miss is configurable still.

A new configuration option was added, `scheduler.runTasksBufferMillis`, which allows making it so mid-tick-tasks stop *early* by a tiny bit to ensure the deadline isn't missed. This is 0.1ms by default. The region also will not attempt to run mid-tick-tasks if the region deadline is within the buffer range.

The configuration description:
```
Buffer time (in milliseconds) before tick deadline to stop executing intermediate tasks. Ensures runTick() can start on time, at the deadline. Higher = safer, lower = more work done.
```

Generally, this rewrite is intended to fix the issues from the beta scheduler, and provide a safe, production-tested alternative(yes this was tested in production) to the beta scheduler that won't royally screw over people with higher player counts because the beta scheduler decides it wants to do something stupid :/

Do note, this also has a noticeable, yet minor, performance improvement when testing in some production and local environments. This scheduler should be safe to merge, but I'm having some production servers test this further *before* merging, since I want to ensure the bugs from the beta scheduler(replicatable in both Canvas and the beta scheduler of Folia) are all fixed up and won't happen again. If the test continues to go well, this should serve as a good quality replacement for the beta scheduler.


Dueris:
After this PR is merged, I also think it is safe to say that we can restore max concurrent players, since that removal was made because of the beta scheduler conflict with it.

R00tB33rMan (commented):
Would it be beneficial to explain why 0.3ms is the suggested base interval?

R00tB33rMan:
Beyond the comment above, everything else is a-okay (conflict tests were conducted as well). I went ahead and deployed these changes pre-merge to verify that all logic is in order with a 300-400 concurrent PB instance with zero discrepancies. I could not observe any newfound exceptions, crashes, & etc. Furthermore, various tests were created locally to confirm that any and all edge cases (specifically pinning) do not create any detrimental issues on local and production servers.

Dueris:
> Would it be beneficial to explain why 0.3ms is the suggested base interval?

I made it 0.3 because the buffer shouldn't be too large where it just started doing intermediate tasks less, and not too small where larger tasks could make the deadline be missed by a tad. It is probably not needed, and can be set really small, however I found this number to be a nice balance between "i need to run tasks" and "i need to be on time". Smaller numbers may work fine, I just did 0.3ms to be safe when originally creating it.

R00tB33rMan (approved):

coredex-source (approved):

coredex-source:
Seems fine, cleared my doubts with @Dueris on discord.

Dueris:
ATM all tests are running perfectly. I am comfortable with merging this.

Dueris:
Issues with the beta scheduler also seem to be resolved.

Toffikk (approved):

PR #148, opened 2025-12-28
Fix ender pearls despawning when switching dimensions

coredex-source:
- Swapped to a better fix that does not remove pearl ticking.
- !entity.isAlive() caused it to despawn when swapping dimensions.
- entity.getBukkitEntity().taskScheduler.isRetired() breaks the gamerule and causes a pearl not disappearing on disconnecting from the server.
- serverPlayer.deathTime > 0 patches it by checking the time of death instead to patch the issue while maintaining parity.


coredex-source:
Nono bad patch.

PR #152, opened 2026-01-03
feat: update for weaver auto AT file discovery for non-minecraft sources

Toffikk:


Toffikk:
don't merge until weaver 2.3.12 is released to avoid issues with jenkins

PR #154, opened 2026-01-06
Secure 1024-bit seed feature generation

Chesvin1:
This ports Luminol (Leaf)'s secure feature seed implementation. Added a setting to enable feature/structure RNG that uses 1024‑bit seed while terrain/biomes remain tied to the normal 64‑bit world seed. 
This is useful to prevent seed cracking.

- Added `"secureSeed": false` setting to the config.
- `/seed` shows 1024-bit feature seed when secure seed is enabled.

- Original patch implementation is by @MrHua269 in Luminol, feature is originally from Leaf.

### DIFF
--- a/canvas-server/minecraft-patches/base/0008-Leaf-Secure-seed-and-matter-seed-command.patch
+++ b/canvas-server/minecraft-patches/base/0008-Leaf-Secure-seed-and-matter-seed-command.patch
@@ -1,0 +1,419 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: MrHua269 <wangxyper@163.com>
+Date: Tue, 28 Jan 2025 18:56:53 +0800
+Subject: [PATCH] Leaf: Secure seed and matter seed command
+
+Ported to Canvas by: Chesvin1 <Chesvin1@hotmail.com>
+
+diff --git a/net/minecraft/server/commands/SeedCommand.java b/net/minecraft/server/commands/SeedCommand.java
+index 7c1e18d8362be5ae885c32b05e98b9ef45942d93..ae2a52193eb87467c31ed79b95e15d9b50f56076 100644
+--- a/net/minecraft/server/commands/SeedCommand.java
++++ b/net/minecraft/server/commands/SeedCommand.java
+@@ -12,6 +12,16 @@ public class SeedCommand {
+             long seed = commandContext.getSource().getLevel().getSeed();
+             Component component = ComponentUtils.copyOnClickText(String.valueOf(seed));
+             commandContext.getSource().sendSuccess(() -> Component.translatable("commands.seed.success", component), false);
++            // Leaf start - Matter - SecureSeed Command
++            if (io.canvasmc.canvas.Config.INSTANCE.secureSeed) {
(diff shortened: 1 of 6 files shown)


Toffikk:
#123 
> Im not entirely comfortable adding secure seed into Canvas currently, as from what Ive seen is that this option breaks world gen datapacks, and for people moving from Folia -> Canvas, those datapacks are often times the only way to add those worlds. By implementing secure seed, we risk people enabling this thinking "oh look fancy security patch" when it will just break their worlds. If this were to become stable && not break world gen datapacks, I would be more inclined to adding this. However for the time being I do not wish to implement this.

Chesvin1:
> #123
> 
> > Im not entirely comfortable adding secure seed into Canvas currently, as from what Ive seen is that this option breaks world gen datapacks, and for people moving from Folia -> Canvas, those datapacks are often times the only way to add those worlds. By implementing secure seed, we risk people enabling this thinking "oh look fancy security patch" when it will just break their worlds. If this were to become stable && not break world gen datapacks, I would be more inclined to adding this. However for the time being I do not wish to implement this.

I use this with world generation datapacks, it doesn't break them. I have tested it with big world generation datapacks such as Terralith and Incendium, and it works just fine. I actually used Folia in the past and switched to the 1024 bit feature generation without issues. 
As mentioned above, this doesn't mess with biome or terrain generation, only feature generation.
I tested it before doing this PR, and I can confirm world generation datapacks work, including datapacks that add new features/structures.

Dueris:
I will do further testing of this PR later and a proper review. Thank you for ur submission, Ill try and get it done sometime today

Dueris:
This PR, the Luminol patch, it is *fully* based off the Leaf patch? I am asking because I did find a discussion in the leaf discord that regards this patch that does alter world generation, so I am currently unsure about this given currently it is said that the issue in reference happens on all versions of Leaf, meaning if Luminols is a 1:1 port, it most likely has this issue too.

https://discord.com/channels/1145991395388162119/1379560713642704967

Chesvin1:
It is, yes.
Hm, I honestly never noticed those straight lines along biome borders, I had to do a triple check in my server and now I notice it in some places 🤔. As far as I can tell, it doesn’t seem to have any impact beyond that (and it's not really noticeable at all at ground level). We’ve been running our server with a 200k × 200k world using 1024-bit feature generation for quite a while, and I’ve never really seen it come up. Players haven’t mentioned it either, so I’m not sure if it’s a significant issue.

To me, it seems like a reasonable compromise in exchange for a secure seed (no way for hacked clients to bypass antixray for example, by knowing the server's seed)

My suggestion would be to add this option disabled by default, along with a disclaimer that it may cause minor visual changes to biome borders.... 
I’ll dig into it a bit more though, and see if I can figure out why that happens, since as far as I know this only affects feature generation.

Dueris:
Yea ATM, despite it being a small issue, I would infact like this issue to be resolved, I think Leaf and Luminol would too.

If this issue can be resolved, Ill probably merge the PR

Dueris:
Until this issue is resolved, Im going to close up this PR, as ATM we have no plans on merging this, and I am not comfortable with merging this until all issues with this are resolved.

PR #155, opened 2026-01-28
Remove JDs

Dueris:
Authored by Toffik, removes JD provider from Canvas source


R00tB33rMan (approved):
Nothing that I can find here that's breaking! LGTM

PR #164, opened 2026-02-22
Fix large cross-region single-tick movement by rerouting to async teleport

BaconCat1:
Prevent cross-region concurrent modification from extreme single-tick movement (for example, pearl guns launching entities across thousands of blocks).
In Entity#move, detect large movement (>= 30 blocks in one tick) that is not fully owned by the current tick region.
For eligible entities (non-vehicle, non-passenger), reroute movement to teleportAsync(..., TELEPORT_FLAG_LOAD_CHUNK) and return early on success.

This is important because a very large movement in a single tick can cross region boundaries and cause unsafe cross-region access. 

A similar fix has also been added to Luminol. 
https://github.com/LuminolMC/Luminol/commit/2616bd9df6270a232cd424f84dae9715a32088a8#diff-c4059dd53771ddb7a2352f44a3783e957ba495a0d4d910dc4b16639ea4787f83R1-R80
However it had several issues which have also been fixed:
- Could get entities stuck (preventMoving not always cleared).
- Teleport call could fail for passengers/vehicles.
- Checked only destination region, not full path.
- Triggered too broadly (not just huge jumps).
- Used unsafe ServerLevel cast.

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
@@ -68,7 +68,7 @@
                  Vec3 vec32 = this.position();
                  Vec3 vec33 = vec32.add(vec3);
 +                // Canvas start - safely reroute very large cross-region movement
-+                if (vec3.lengthSqr() >= (32.0D * 32.0D)
++                if (vec3.lengthSqr() >= (30.0D * 30.0D)
 +                    && !this.isVehicle()
 +                    && !this.isPassenger()
 +                    && !ca.spottedleaf.moonrise.common.util.TickThread.isTickThreadFor(this.level(), vec32, vec3, 0)
--- a/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
@@ -63,12 +63,34 @@
          }
  
          this.firstTick = false;
+@@ -1115,6 +_,21 @@
+         final Vec3 originalMovement = movement; // Paper - Expose pre-collision velocity
+         ca.spottedleaf.moonrise.common.util.TickThread.ensureTickThread("Cannot move an entity off-main"); // Folia - region threading
+         if (this.noPhysics) {
++            // Canvas start - safely reroute very large cross-region movement
++            Vec3 vec32 = this.position();
++            if (movement.lengthSqr() >= (30.0D * 30.0D)
++                && !this.isVehicle()
++                && !this.isPassenger()
++                && !ca.spottedleaf.moonrise.common.util.TickThread.isTickThreadFor(this.level(), vec32, movement, 0)
++                && this.level() instanceof ServerLevel serverLevel
++                && this.teleportAsync(
++                serverLevel, vec32.add(movement), null, null, this.getDeltaMovement(),
++                org.bukkit.event.player.PlayerTeleportEvent.TeleportCause.UNKNOWN,
++                TELEPORT_FLAG_LOAD_CHUNK, null
++            )) {
++                return;
++            }
++            // Canvas end - safely reroute very large cross-region movement
+             this.setPos(this.getX() + movement.x, this.getY() + movement.y, this.getZ() + movement.z);
+             this.horizontalCollision = false;
+             this.verticalCollision = false;
 @@ -1163,6 +_,20 @@
  
                  Vec3 vec32 = this.position();
                  Vec3 vec33 = vec32.add(vec3);
 +                // Canvas start - safely reroute very large cross-region movement
-+                if (vec3.lengthSqr() >= (32.0D * 32.0D)
++                if (vec3.lengthSqr() >= (30.0D * 30.0D)
 +                    && !this.isVehicle()
 +                    && !this.isPassenger()
 +                    && !ca.spottedleaf.moonrise.common.util.TickThread.isTickThreadFor(this.level(), vec32, vec3, 0)
@@ -275,8 +297,8 @@
 +        }
 +        return this.calculateViewVector(this.canvas$threadSafeXRot, this.canvas$threadSafeYRot);
 +    }
-+
 +    // Canvas end - region threading
++
      public Vec3 getHeadLookAngle() {
          return this.calculateViewVector(this.getXRot(), this.getYHeadRot());
      }
--- a/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
@@ -68,7 +68,7 @@
                  Vec3 vec32 = this.position();
                  Vec3 vec33 = vec32.add(vec3);
 +                // Canvas start - safely reroute very large cross-region movement
-+                if (vec3.lengthSqr() >= (128.0D * 128.0D)
++                if (vec3.lengthSqr() >= (32.0D * 32.0D)
 +                    && !this.isVehicle()
 +                    && !this.isPassenger()
 +                    && !ca.spottedleaf.moonrise.common.util.TickThread.isTickThreadFor(this.level(), vec32, vec3, 0)
--- a/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
@@ -63,49 +63,40 @@
          }
  
          this.firstTick = false;
-@@ -1115,6 +_,21 @@
+@@ -1115,6 +_,17 @@
          final Vec3 originalMovement = movement; // Paper - Expose pre-collision velocity
          ca.spottedleaf.moonrise.common.util.TickThread.ensureTickThread("Cannot move an entity off-main"); // Folia - region threading
          if (this.noPhysics) {
-+            // Canvas start - safely reroute very large cross-region movement
++            // Canvas start - safely abort tick when very large movement crosses region ownership
 +            Vec3 vec32 = this.position();
++            Vec3 vec33 = vec32.add(movement);
 +            if (movement.lengthSqr() >= (30.0D * 30.0D)
++                && !Double.isNaN(vec33.x) && !Double.isNaN(vec33.y) && !Double.isNaN(vec33.z)
 +                && !this.isVehicle()
 +                && !this.isPassenger()
-+                && !ca.spottedleaf.moonrise.common.util.TickThread.isTickThreadFor(this.level(), vec32, movement, 0)
-+                && this.level() instanceof ServerLevel serverLevel
-+                && this.teleportAsync(
-+                serverLevel, vec32.add(movement), null, null, this.getDeltaMovement(),
-+                org.bukkit.event.player.PlayerTeleportEvent.TeleportCause.UNKNOWN,
-+                TELEPORT_FLAG_LOAD_CHUNK, null
-+            )) {
-+                return;
++                && !ca.spottedleaf.moonrise.common.util.TickThread.isTickThreadFor(this.level(), vec32, movement, 0)) {
++                throw new io.canvasmc.canvas.entity.EntityMoveOutOfRegionException(this, type, movement);
 +            }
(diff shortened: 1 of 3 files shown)


Dueris:
Would also like it to be known that in the server, casting to ServerLevel is always safe if its a Level from an entity or player or something, since its the server side, not including the client side

BaconCat1:
> Would also like it to be known that in the server, casting to ServerLevel is always safe if its a Level from an entity or player or something, since its the server side, not including the client side

Oh yeah that’s a good point.

Dueris:
I also am unsure how I feel about a specific block count that needs to be exceeded before we try checking, in this case 128 blocks. Why choose that? And what happens if say, the render distance for the server is like... 2?

Dueris:
I like having a block count needing to be exceeded, since it prevents extreme spam of isTickThreadFor calls for every entity move, I just wanna know why 128 and if we should make this smaller, say 30+ blocks in 1 movement operation

BaconCat1:
> I also am unsure how I feel about a specific block count that needs to be exceeded before we try checking, in this case 128 blocks. Why choose that? And what happens if say, the render distance for the server is like... 2?

I chose 128 block because it’s 8 chunks or the vanilla simulation distance default, but in hindsight we should be determining it based on render distance.

Dueris:
Agreed. Or we just limit it to like 30 blocks, since realistically u won't be going bellow 2 chunk render distances(and I dont even think you can), and that's 32 blocks, but you also include things like protochunks and such outside of render distance, plus the region exponent, and 30 is a pretty damn good number

canvas-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
MrMasrozYTLIVE, line 1: Lgtm

MrMasrozYTLIVE (approved):

Dueris (changes requested):
Id like there to be a configuration option

Dueris:
Im taking a tiny bit to review this, trying to make sure everything is all safe and sound. Im also curious if we need a config, or if we should keep this without an option to match the rest of the fixes we have to Folia being natively enabled

BaconCat1:
> Im taking a tiny bit to review this, trying to make sure everything is all safe and sound. Im also curious if we need a config, or if we should keep this without an option to match the rest of the fixes we have to Folia being natively enabled

I was debating this and decided against a config option since it’s a safety fix and many other canvas changes to folia do not have a config. I can add one, but do you think it should be on or off by default?

canvas-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
```diff
@@ -63,6 +63,27 @@
          }
  
          this.firstTick = false;
+@@ -1163,6 +_,20 @@
+ 
+                 Vec3 vec32 = this.position();
+                 Vec3 vec33 = vec32.add(vec3);
```
Dueris, line 87: What I am a bit confused about is why this is placed here. Realistically, this isn't the best place at all to inject this, due to the following code.
```java
public void move(MoverType type, Vec3 movement) {
        final Vec3 originalMovement = movement; // Paper - Expose pre-collision velocity
        ca.spottedleaf.moonrise.common.util.TickThread.ensureTickThread("Cannot move an entity off-main"); // Folia - region threading
        if (this.noPhysics) {
            this.setPos(this.getX() + movement.x, this.getY() + movement.y, this.getZ() + movement.z);
            this.horizontalCollision = false;
            this.verticalCollision = false;
            this.verticalCollisionBelow = false;
            this.minorHorizontalCollision = false;
        } else {
```
At the head, we check it's the same tick thread, yippe. But if the entity has no physics, it just sets the position up at the head of the method. So, Im curious if we should move this to the HEAD of the method after the thread check, and then use the `movement` var passed to check this, since the local var `vec3`, is checked using `collide(movement)`, which I am not sure if that's cross-region safe, assuming the circumstances where this code would be executed are being utilized.

Dueris (changes requested):

Dueris:
> I was debating this and decided against a config option since it’s a safety fix and many other canvas changes to folia do not have a config. I can add one, but do you think it should be on or off by default?

On by default fs, and I agree, safety fixes Canvas doesn't commonly add configs for. So Id say its fine to leave as-is in terms of configurations

canvas-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
BaconCat1: We do this after collide(movement) on purpose, because vec3 is the actual resolved movement that the game will apply. If it's at the HEAD and using raw movement may result in teleporting based on un-collided motion, bypassing normal collision resolution. The main unsafe part we are mitigating is the cross region apply step setPos. noPhysics is a fair edge case callout, I currently just targeted normal physics movement (the pearl-gun scenario).

BaconCat1: But you are right collide might not be thread safe

Dueris: If `collide(movement)` isn't thread-safe though, we don't actually fix anything.

BaconCat1: I've just looked into it. `collide()` itself is region guarded (non-owned chunks are treated as blocking), so the main unsafe part is still the final apply step, which this fixes, hooray!

Dueris: Aight then. However I would probably still like it if we could have a similar check at the HEAD for non-physics entities. While probably not gonna be encountered in Survival, Id like that to be covered for correctness sake.

Dueris: But yea that is very interesting and makes sense that it would be considered as blocking, thanks for looking into that, currently a bit busy irl so didn't have enough time to fully look into that but wanted to run by if that was something to be concerned about

BaconCat1: Yeah, I was thinking of this too.

Dueris:
Id like a comment in the patch saying why we do it there, specifically also explain collide having unloaded chunks function as blocking

After that, LGTM

BaconCat1:
> Id like a comment in the patch saying why we do it there, specifically also explain collide having unloaded chunks function as blocking
> 
> After that, LGTM

Done!

Dueris:
And this is production tested?

R00tB33rMan (approved):
I saw this briefly in its early stages and had a feeling it needed a little time to sit. Thanks for working with us — pretty sure this is good now!

BaconCat1:
> And this is production tested?

For the last week yeah

Dueris:
Out of curiosity does the move event get called in there?

BaconCat1:
> Out of curiosity does the move event get called in there?

Sometimes

BaconCat1:
> > Out of curiosity does the move event get called in there?
> 
> Sometimes

And also something very similar has been tested a lot by Luminol

Dueris:
Yea I am curious about if we should take a similar route and throw a new throwable if it's out of region, and then we catch it during the entity tick and if it does catch it then we schedule to teleport async, like the current Luminol patch. That way we don't create random issues like if the entity tick were to touch something it shouldn't after the move

coredex-source (approved):

BaconCat1:
> Yea I am curious about if we should take a similar route and throw a new throwable if it's out of region, and then we catch it during the entity tick and if it does catch it then we schedule to teleport async, like the current Luminol patch. That way we don't create random issues like if the entity tick were to touch something it shouldn't after the move

I wouldn't use a throwable for this, because we already guard out of region movement in `move` and return immediately after scheduling teleportAsync so the normal post movement logic for that path does not continue in the same tick. Using Throwables would have much higher overhead. Additionally, if something were to actually happen it complicates debugging/stack traces. 

If we still want extra safety we could set a short lived "movement-deferred" flag and skip any move-dependent side effects for that tick.

Dueris:
Well, I meant similar to this:
https://github.com/LuminolMC/Luminol/blob/ver/1.21.11/luminol-server/minecraft-patches/features/0064-Teleport-async-if-entity-was-moving-to-another-regio.patch

Where it's a custom throwable that when caught, teleports. It's very similar to how portals and such work, and is probably significantly safer given the rest of the entity tick logic can do basically anything(especially when you count plugins)

BaconCat1:
> Well, I meant similar to this:
> 
> https://github.com/LuminolMC/Luminol/blob/ver/1.21.11/luminol-server/minecraft-patches/features/0064-Teleport-async-if-entity-was-moving-to-another-regio.patch
> 
> 
> 
> Where it's a custom throwable that when caught, teleports. It's very similar to how portals and such work, and is probably significantly safer given the rest of the entity tick logic can do basically anything(especially when you count plugins)

That's a valid point I'll look into this later - kinda forgot plugins exist and can mess stuff up.

BaconCat1:
Something like this??? Now the issue with this change is that it has not been tested lol, so maybe hold off on merging this until I can properly test it and also needs re-review.

canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
```diff
@@ -68,6 +68,38 @@
...
          regionizedWorldData.seTtickingBlockEntities(false); // Folia - regionised ticking
          regionizedWorldData.currentPrimedTnt = 0; // Spigot // Folia - region threading
+@@ -1463,6 +_,31 @@
+     public <T extends Entity> void guardEntityTick(Consumer<T> action, T entity) {
+         try {
+             action.accept(entity);
++        } catch (io.canvasmc.canvas.entity.EntityMoveOutOfRegionException moveOutOfRegionException) {
++            // Canvas start - reroute cross-region movement after tick unwind
++            final net.minecraft.world.entity.Entity movingEntity = moveOutOfRegionException.entity();
++            if (movingEntity.isRemoved()) {
```
Dueris: Why this?

canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
```diff
@@ -68,6 +68,38 @@
  
          regionizedWorldData.seTtickingBlockEntities(false); // Folia - regionised ticking
          regionizedWorldData.currentPrimedTnt = 0; // Spigot // Folia - region threading
+@@ -1463,6 +_,31 @@
+     public <T extends Entity> void guardEntityTick(Consumer<T> action, T entity) {
+         try {
+             action.accept(entity);
++        } catch (io.canvasmc.canvas.entity.EntityMoveOutOfRegionException moveOutOfRegionException) {
++            // Canvas start - reroute cross-region movement after tick unwind
```
Dueris: Should be moved to encompass the start of the catch block

canvas-server/src/main/java/io/canvasmc/canvas/entity/EntityMoveOutOfRegionException.java
```diff
@@ -0,0 +1,34 @@
+package io.canvasmc.canvas.entity;
+
+public final class EntityMoveOutOfRegionException extends RuntimeException {
```
Dueris, line 7: You can add imports, it's Canvas src here

Dueris (changes requested):
Generally this is what I saw from quickly looking on my phone, Ill do a more thorough review tomorrow morning.

canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
BaconCat1: I initially added it to avoid scheduling a teleport for an already removed entity, but in hindsight the scheduler/teleport path already handles entity lifecycle safely, so it's not actually needed here.

Dueris:
Awaiting results of production testing, LMK once that is done and then I think I am good with merging this. Ill do one last check sometime today, just incase further changes and/or fixes are made, and so that everything is caught and nitpicks are taken care of.

BaconCat1:
Alright, I'm happy with the testing of this and all seems well. Didn't find any problems during testing.

Dueris (approved):

PR #166, opened 2026-02-24
Add setter for destination world in EntityPortalAsyncEvent

ammodev:
This pull request updates the `EntityPortalAsyncEvent` class to allow modification of the destination world for an entity portal event. The most important changes are:

**Event mutability improvements:**

* Changed the `to` field in `EntityPortalAsyncEvent` from `final` to mutable, allowing it to be updated after event creation.
* Added a new `setTo(World to)` method to allow setting the destination world for the portal event.


Dueris:
Nice catch!

PR #170, opened 2026-03-04
Affinity Scheduler

Dueris:
This adds the `AFFINITY` scheduler to Canvas, replacing the old `CRS`, now `StealingQueueSchedulerThreadPool`.

This scheduler adds per tick thread CPU affinity, which can be optionally disabled. It is based on the `EDF` scheduler, and also supports mid-tick-tasks like normal.

Work stealing works as:
  - peek global and local head
  - if local head is overdue, return that
  - if global head is overdue, return that
  - if still here, figure out the soonest deadline between the local and global head
    - if tied, it should be fine, since the comparator compares the IDs which are always unique per task, however we prefer the local task if it does happen
  - increment the next thread we will steal from. we try 1 steal per poll if we reach here, and for each poll we increment through the runners in order
  - if the thread we are trying to steal from isn't the current runner, we peek at the queue, and if it is stealable we return that, since it is past the steal thres
  - if we didn't return the potential stealable, we return the local head or global head depending on which has the soonest deadline, or if both are null, we return null

This also heavily cleans up the region profiler, which has been needed for quite a while

### DIFF
--- a/canvas-server/src/main/java/io/canvasmc/canvas/spark/profiler/RegionScheduleHandlePinner.java
+++ b/canvas-server/src/main/java/io/canvasmc/canvas/spark/profiler/RegionScheduleHandlePinner.java
@@ -207,9 +207,9 @@ public interface RegionScheduleHandlePinner {
         public void unpin(Consumer<SchedulableTick> finalizer) {
             final long[] curr = gatherChunksToProfile();
 
-            long packed = curr[0];
-            int chunkX = (int) packed;
-            int chunkZ = (int) (packed >> 32);
+            final ChunkPos center = getCenter();
+            final int chunkX = center.x;
+            final int chunkZ = center.z;
 
             // Note: this should be loaded already, however we run this safe version just to be sure
             world.canvas$loadOrRunAtChunksAsync(
--- a/canvas-server/src/main/java/io/canvasmc/canvas/spark/profiler/RegionScheduleHandlePinner.java
+++ b/canvas-server/src/main/java/io/canvasmc/canvas/spark/profiler/RegionScheduleHandlePinner.java
@@ -2,26 +2,30 @@ package io.canvasmc.canvas.spark.profiler;
 
 import ca.spottedleaf.concurrentutil.scheduler.SchedulableTick;
 import ca.spottedleaf.concurrentutil.util.Priority;
+import ca.spottedleaf.moonrise.common.util.CoordinateUtils;
+import ca.spottedleaf.moonrise.common.util.TickThread;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import com.mojang.brigadier.exceptions.Dynamic2CommandExceptionType;
 import com.mojang.brigadier.exceptions.SimpleCommandExceptionType;
-import io.canvasmc.canvas.tick.CRSThreadPool;
-import io.canvasmc.canvas.util.COWLongArrayList;
+import io.canvasmc.canvas.tick.AffinitySchedulerThreadPool;
 import io.papermc.paper.threadedregions.RegionizedServer;
 import io.papermc.paper.threadedregions.ThreadedRegionizer;
 import io.papermc.paper.threadedregions.TickRegionScheduler;
 import io.papermc.paper.threadedregions.TickRegions;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import it.unimi.dsi.fastutil.longs.LongArrayList;
+import it.unimi.dsi.fastutil.longs.LongComparator;
 import net.minecraft.core.BlockPos;
 import net.minecraft.server.level.ChunkMap;
 import net.minecraft.server.level.ColumnPos;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.server.level.Ticket;
 import net.minecraft.server.level.TicketType;
+import net.minecraft.world.level.ChunkPos;
 import net.minecraft.world.level.border.WorldBorder;
 import org.jetbrains.annotations.NotNull;
+import org.jspecify.annotations.NonNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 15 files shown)


Toffikk (approved):
overall LGTM

yunuservices (approved):
absolute LGTM, good

canvas-server/src/main/java/io/canvasmc/canvas/spark/profiler/RegionScheduleHandlePinner.java
```diff
@@ -199,9 +248,9 @@ public Logger getLogger() {
      */
     class GlobalTickPinner implements RegionScheduleHandlePinner {
         @Override
-        public void pin(BiConsumer<TickRegionScheduler.RegionScheduleHandle, CRSThreadPool.TickThreadRunner> finalizer, CRSThreadPool crsThreadPool) throws CommandSyntaxException {
+        public void pin(BiConsumer<TickRegionScheduler.RegionScheduleHandle, AffinitySchedulerThreadPool.TickThreadRunner> finalizer) throws CommandSyntaxException {
             TickRegionScheduler.RegionScheduleHandle schedulingHandle = RegionizedServer.getGlobalTickData();
-            final CRSThreadPool.TickThreadRunner thread = ((CRSThreadPool.ScheduledState) schedulingHandle.state).getOwnedBy();
+            final AffinitySchedulerThreadPool.TickThreadRunner thread = ((AffinitySchedulerThreadPool) TickRegions.getScheduler().scheduler).getCurrentTickThreadRunner();
```
BaconCat1, line 253: Global tick profiling might be broken by the new runner lookup. `GlobalTickPinner.pin` calls `AffinitySchedulerThreadPool.getCurrentTickThreadRunner()` from the global tick thread, but that helper only succeeds on region scheduler worker threads. The same assumption is repeated during unpin in `SparkRegionProfilerExtension.endPinning`, meaning `/spark profiler --thread * --region global` will fail both when starting and when stopping.

canvas-server/src/main/java/io/canvasmc/canvas/tick/SchedulerUtil.java
```diff
@@ -120,83 +127,70 @@ public void onRegionInactive(final ThreadedRegionizer.@NonNull ThreadedRegion<Ti
...
         }
 
         @Override
-        public void tryTransferPinningState(final TickRegionScheduler.@NonNull RegionScheduleHandle ret, final TickRegionScheduler.RegionScheduleHandle to) {
-            if (ret.state instanceof CRSThreadPool.ScheduledState crsState)
-                crsState.pin(((CRSThreadPool.ScheduledState) to.state).getPinnedThreadId(), crsState.schedulerOwnedBy);
+        public void tryTransferPinningState(final TickRegionScheduler.@NonNull RegionScheduleHandle from, final TickRegionScheduler.RegionScheduleHandle to) {
+            if (!isRunningRegionProfiler() || !(TickRegions.getScheduler().scheduler instanceof AffinitySchedulerThreadPool affinitySchedulerThreadPool)) {
+                return;
+            }
```
BaconCat1, line 150: Should from/to be reversed here? The old CRS version copied the pin state from the currently pinned handle onto the replacement handle, but this implementation checks isLinkedTo(from) before relinking. With the current call sites, I think from may actually be the new handle, which would skip the transfer entirely. Or is this intentional?

canvas-server/src/main/java/io/canvasmc/canvas/spark/profiler/RegionScheduleHandlePinner.java
```diff
@@ -86,11 +90,55 @@ public interface RegionScheduleHandlePinner {
      *     the world we are operating on
      */
     record RegionPinner(ColumnPos fromPos, ColumnPos toPos, ServerLevel world) implements RegionScheduleHandlePinner {
-        public static final COWLongArrayList PROFILING_CHUNKS = new COWLongArrayList();
-        public static final AtomicReference<ServerLevel> PROFILING_LEVEL = new AtomicReference<>();
+        public @NonNull ChunkPos getCenter() {
```
BaconCat1, line 93: Split tracking and stop profiling cleanup no longer target the same child region for multi-chunk profiles. On split, it will keeps the profiler on the region containing `RegionPinner.getCenter()` via `SchedulerUtil.onRegionSplit`. But `RegionPinner.unpin()` still schedules cleanup through curr[0], the first chunk in the selection. After a split, those can be different child regions, so endPinning may unlink the wrong runner and leave the actual profiled runner linked.

canvas-server/src/main/java/io/canvasmc/canvas/tick/SchedulerUtil.java
Dueris: Unintentional, was called wrong in TickRegions

canvas-server/src/main/java/io/canvasmc/canvas/spark/profiler/RegionScheduleHandlePinner.java
Dueris: All chunks in the profile are forceloaded, it doesn't entirely matter if we target any chunk currently profiled, its guaranteed to be correct because it is all forceloaded to the same region. I do wanna change this to the center for simplicity and consistency tho, so Ill keep this marked unresolved so I am reminded to do that later

canvas-server/src/main/java/io/canvasmc/canvas/spark/profiler/RegionScheduleHandlePinner.java
Dueris: Why would it not succeed? It works locally, and the global tick is run on region scheduler threads

BaconCat1: If global tick is guaranteed to always execute on an AffinitySchedulerThreadPool runner, then this is probably fine.

My concern was just that getCurrentTickThreadRunner() only succeeds if Thread.currentThread() is one of the scheduler runners, and the old CRS code looked up the owner from the handle state instead of assuming that.

canvas-server/src/main/java/io/canvasmc/canvas/spark/profiler/RegionScheduleHandlePinner.java
BaconCat1: Sounds good!

Dueris:
Should be finished, ready for full review

R00tB33rMan (approved):
Approval for testing!

Edit: Tested successfully with 850+ players and no crashes for 22 hours!

R00tB33rMan:
Testing is completed! No issues were found.

Dueris:
I plan on merging this in 6 hours or so. If anyone, core team or not, wants to do a final review, feel free.

PR #173, opened 2026-03-20
Cache Level#getEntities results by bounding box

BaconCat1:
Adds an optional per-tick cache for Level#getEntities, keyed by bounding box, to reduce repeated lookups.
Stores results in a regionised cache, cleared each tick.

In testing with ~1000 cows cramming in a small 6x6 space it was around 5x faster than vanilla.

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
@@ -1,5 +1,30 @@
 --- a/net/minecraft/world/level/Level.java
 +++ b/net/minecraft/world/level/Level.java
+@@ -107,6 +_,24 @@
+     public static final int MAX_BRIGHTNESS = 15;
+     public static final int MAX_ENTITY_SPAWN_Y = 20000000;
+     public static final int MIN_ENTITY_SPAWN_Y = -20000000;
++    // Canvas start - cache getEntities lookups by bounding box for the current tick
++    private static final io.papermc.paper.threadedregions.RegionizedData<Level.CanvasGetEntitiesBoxCache> canvas$regionizedGetEntitiesBoxCache =
++        new io.papermc.paper.threadedregions.RegionizedData<>(null, regionData -> new Level.CanvasGetEntitiesBoxCache(), new io.papermc.paper.threadedregions.RegionizedData.RegioniserCallback<>() {
++            @Override
++            public void merge(final Level.CanvasGetEntitiesBoxCache from, final Level.CanvasGetEntitiesBoxCache into, final long fromTickOffset) {
++                into.clear();
++            }
++
++            @Override
++            public void split(final Level.CanvasGetEntitiesBoxCache from, final int chunkToRegionShift,
++                              final it.unimi.dsi.fastutil.longs.Long2ReferenceOpenHashMap<Level.CanvasGetEntitiesBoxCache> regionToData,
++                              final it.unimi.dsi.fastutil.objects.ReferenceOpenHashSet<Level.CanvasGetEntitiesBoxCache> dataSet) {
++                for (final Level.CanvasGetEntitiesBoxCache cache : dataSet) {
++                    cache.clear();
++                }
++            }
++        });
++    // Canvas end - cache getEntities lookups by bounding box for the current tick
+     public static final WeightedList<ExplosionParticleInfo> DEFAULT_EXPLOSION_BLOCK_PARTICLES = WeightedList.<ExplosionParticleInfo>builder()
+         .add(new ExplosionParticleInfo(ParticleTypes.POOF, 0.5F, 1.0F))
+         .add(new ExplosionParticleInfo(ParticleTypes.SMOKE, 1.0F, 1.0F))
 @@ -203,7 +_,7 @@
  
      @Override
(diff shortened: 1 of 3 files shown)


Dueris:
Im not quite sure I like that it doesn't also check the predicate, since it could lead to missleading results, which may break in-game mechanics, and also cause crashes

BaconCat1:
> Im not quite sure I like that it doesn't also check the predicate, since it could lead to missleading results, which may break in-game mechanics, and also cause crashes

That’s a fair point. I was aware that not including entity/predicate would improve hit rate, and that was part of why I initially kept the key narrow, but I agree it makes the cache too unsafe.
I’ve changed the key to include both the predicate and entity.

canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
```diff
@@ -95,6 +120,77 @@
...
++            cache.clearIfStale();
++
++            final List<Entity> cached = cache.entitiesByBoundingBox.get(Level.canvas$getEntitiesBoxCacheKey(entity, boundingBox, predicate));
++            if (cached != null) {
++                return cached;
++            }
++        }
++
+         final List<Entity> ret = new java.util.ArrayList<>();
+-
```
Dueris: Unnecessary whitespace removal

canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
```diff
@@ -95,6 +120,77 @@
...
+         ((ca.spottedleaf.moonrise.patches.chunk_system.level.ChunkSystemLevel)this).moonrise$getEntityLookup().getEntities(entity, boundingBox, ret, predicate);
+-
```
Dueris: Unnecessary whitespace removal

canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
```diff
@@ -95,6 +120,77 @@
...
+         ca.spottedleaf.moonrise.common.PlatformHooks.get().addToGetEntities((Level)(Object)this, entity, boundingBox, predicate, ret);
+ 
++        if (io.canvasmc.canvas.Config.INSTANCE.useBoxCacheForGetEntities) {
++            Level.canvas$regionizedGetEntitiesBoxCache.get().entitiesByBoundingBox.put(Level.canvas$getEntitiesBoxCacheKey(entity, boundingBox, predicate), ret);
++        }
++        // Canvas end - cache getEntities lookups by bounding box for the current tick
++
```
Dueris: Unnecessary whitespace addition, also undocumented

canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
```diff
@@ -95,6 +120,77 @@
...
++        if (io.canvasmc.canvas.Config.INSTANCE.useBoxCacheForGetEntities) {
++            final Level.CanvasGetEntitiesBoxCache cache = Level.canvas$regionizedGetEntitiesBoxCache.get();
++            cache.clearIfStale();
++
++            final List<Entity> cached = cache.entitiesByBoundingBox.get(Level.canvas$getEntitiesBoxCacheKey(entity, boundingBox, predicate));
++            if (cached != null) {
++                return cached;
++            }
++        }
++
```
Dueris: Unnecessary whitespace addition, also undocumented

canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
```diff
@@ -1,5 +1,30 @@
 --- a/net/minecraft/world/level/Level.java
 +++ b/net/minecraft/world/level/Level.java
+@@ -107,6 +_,24 @@
+     public static final int MAX_BRIGHTNESS = 15;
+     public static final int MAX_ENTITY_SPAWN_Y = 20000000;
+     public static final int MIN_ENTITY_SPAWN_Y = -20000000;
++    // Canvas start - cache getEntities lookups by bounding box for the current tick
```
Dueris: Why did you choose to make separate regionized data, instead of introducing into the `RegionizedWorldData`?

canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
```diff
@@ -95,6 +120,77 @@
...
++            }
++        }
++
++        private void clear() {
++            this.tick = Long.MIN_VALUE;
++            this.entitiesByBoundingBox.clear();
++        }
++    }
++    // Canvas end - cache getEntities lookups by bounding box for the current tick
++
```
Dueris: Undocumented whitespace addition

canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
```diff
@@ -95,6 +120,77 @@
...
+@@ -1719,16 +_,68 @@
+         List<Entity> list = Lists.newArrayList();
+ 
+         // Paper start - rewrite chunk system
++        // Canvas start - cache getEntities lookups by bounding box for the current tick
++        if (io.canvasmc.canvas.Config.INSTANCE.useBoxCacheForGetEntities) {
++            final Level.CanvasGetEntitiesBoxCache cache = Level.canvas$regionizedGetEntitiesBoxCache.get();
++            cache.clearIfStale();
++
++            final List<Entity> cached = cache.entitiesByBoundingBox.get(Level.canvas$getEntitiesBoxCacheKey(entity, boundingBox, predicate));
```
Dueris, line 106: What happens if one of the entities that was cached moves, or gets removed or more entities get added?

Dueris (changes requested):

canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
BaconCat1: Mostly for convenience since the cache is only consumed from Level#getEntities, so I initially kept it local instead of adding another field to RegionizedWorldData. After looking at it more closely, RegionizedWorldData is the better ownership because this cache is for a world local region state

canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
BaconCat1: The cache becomes stale for the rest of that tick. 

So within a single tick:

- if a cached entity moves out of the box, the cached result can still include it
- if a new entity moves into the box or is added, the cached result can miss it
- if an entity is removed, the cached result can still contain that reference until the tick ends.

Dueris: Not entirely sure how I feel regarding this, since a lot can happen in a tick because of plugins and things like that. While it probably won't do much, it could still cause issues

BaconCat1: Yeah my concern is if we add all sorts of checks to it the cache will miss so much it's not even worth having. TBH if we just have some aggressive warning on what it does it could be ok.

Dueris: Yea the optimization wouldn't be effective at all if we add all sorts of checks, however with this, and plugins, this could cause instability issues, crashes, and all sorts of other undiagnosable issues, which is why ATM Im kinda against this.

Dueris:
Yea I think Im gonna close this, while a good idea in concept, this has a lot of edge cases that are difficult to cover while still having a performance benefit. While the chances are slim, this isn't quite something I think we should implement because of the potential instability issues.

PR #174, opened 2026-03-20
Restore chunk ticking guard for block entity ticks

BaconCat1:
Restore the chunk ownership check before ticking block entities to prevent cross-region or non-ticking chunk block entity ticks introduced by the block entity iteration optimization, from commit 649047d7.

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/world/level/Level.java.patch
@@ -42,7 +42,7 @@
 +                continue;
 +            }
 +
-+            if (runsNormally && tickingBlockEntity.getPos() != null) { // block entity sleeping
++            if (runsNormally && tickingBlockEntity.getPos() != null && this.shouldTickBlocksAt(tickingBlockEntity.getPos())) { // Canvas - only tick block entities in ticking chunks
                  tickingBlockEntity.tick();
 -                // Paper start - rewrite chunk system
                  if ((++tickedEntities & 7) == 0) {


Dueris:
`shouldTickBlocksAt` is:
```java
public boolean shouldTickBlocksAt(long chunkPos) {
    // Paper start - rewrite chunk system
    final ca.spottedleaf.moonrise.patches.chunk_system.scheduling.NewChunkHolder holder =  this.moonrise$getChunkTaskScheduler().chunkHolderManager.getChunkHolder(chunkPos);
    return holder != null && holder.isTickingReady();
    // Paper end - rewrite chunk system
}
```

`blockEntityTickers`, the local var in the method `Level#tickBlockEntities`, is defined as `RegionizedWorldData#getBlockEntityTickers` on the current region. I removed this check in the commit https://github.com/CraftCanvasMC/Canvas/commit/649047d7766752022b5fa4b1c72499c6d5ec13ad because this check is duplicated inside the actual tick of the block entity ticker(`BoundTickingBlockEntity`)

```java
boolean canvas$isTicking() {
    if (this.canvas$isInsideWorldBorder()) {
        // no need to check 'LevelChunk.this.level.areEntitiesLoaded', as this will
        // always be true, since as a LevelChunk, we are always at the FULL chunk status
        // so obviously we are loaded. we just need to check if we are block ticking now
        return LevelChunk.this.getFullStatus().isOrAfter(FullChunkStatus.BLOCK_TICKING);
    }
    return false;
}
```
The check was removed simply bc it was kinda redundant and not really needed

Dueris:
The check inside the ticker is actually faster/cheaper too, since the previous call would fetch from the world chunk holders map too

BaconCat1:
That makes sense, thanks for the explanation.

So with this change we’d effectively be checking chunk tickability twice: once via shouldTickBlocksAt(...) here and again inside BoundTickingBlockEntity.
 
Since the ticker already enforces BLOCK_TICKING (and is cheaper), we should keep original logic. Closing this.

PR #178, opened 2026-03-25
feat: configuration options for the tpsbar command

MioArchive:
This very simple PR adds the ability to customize the TpsBar command in the canvas config. In addition, I've added the region's player count to the TpsBar (everyone can decide for themselves via the config whether they want to display the player count or not)

### DIFF
--- a/canvas-server/src/main/java/io/canvasmc/canvas/Config.java
+++ b/canvas-server/src/main/java/io/canvasmc/canvas/Config.java
@@ -485,6 +485,16 @@ public class Config {
     })
     public boolean enableTpsBar = true;
 
+    @Comment({
+        "MiniMessage-formatted line for the TPS bar. Placeholders: <tps>, <mspt>, <util>, <players>.",
+        "Legacy tokens %tps%, %mspt%, %util%, %players% are also accepted and auto-converted."
+    })
+    public String tpsBarFormat =
+        "<gradient:#357cef:#f21af4><b>TPS</b></gradient>: <tps>  -  " +
+        "<gradient:#357cef:#f21af4><b>MSPT</b></gradient>: <mspt>  -  " +
+        "<gradient:#357cef:#f21af4><b>Util</b></gradient>: <util>%  -  " +
+        "<gradient:#357cef:#f21af4><b>Players</b></gradient>: <players>";
+
     @Comment(value = {
         "The default respawn dimension for the server.",
         "This can assist for servers that need this changed to a different world",
--- a/canvas-server/src/main/java/io/canvasmc/canvas/Config.java
+++ b/canvas-server/src/main/java/io/canvasmc/canvas/Config.java
@@ -489,11 +489,7 @@ public class Config {
         "MiniMessage-formatted line for the TPS bar. Placeholders: <tps>, <mspt>, <util>, <players>.",
         "Legacy tokens %tps%, %mspt%, %util%, %players% are also accepted and auto-converted."
     })
-    public String tpsBarFormat =
-        "<gradient:blue:aqua><b>TPS</b></gradient>: <tps>  -  " +
-            "<gradient:blue:aqua><b>MSPT</b></gradient>: <mspt>  -  " +
-            "<gradient:blue:aqua><b>Util</b></gradient>: <util>%  -  " +
-            "<gradient:blue:aqua><b>Players</b></gradient>: <players>";
+    public String tpsBarFormat = RegionizedTpsBar.DEFAULT_FORMAT;
 
     @Comment(value = {
         "The default respawn dimension for the server.",
--- a/canvas-server/src/main/java/io/canvasmc/canvas/Config.java
+++ b/canvas-server/src/main/java/io/canvasmc/canvas/Config.java
@@ -490,10 +490,10 @@ public class Config {
         "Legacy tokens %tps%, %mspt%, %util%, %players% are also accepted and auto-converted."
     })
     public String tpsBarFormat =
-        "<gradient:#357cef:#f21af4><b>TPS</b></gradient>: <tps>  -  " +
-        "<gradient:#357cef:#f21af4><b>MSPT</b></gradient>: <mspt>  -  " +
-        "<gradient:#357cef:#f21af4><b>Util</b></gradient>: <util>%  -  " +
-        "<gradient:#357cef:#f21af4><b>Players</b></gradient>: <players>";
+        "<gradient:blue:aqua><b>TPS</b></gradient>: <tps>  -  " +
+            "<gradient:blue:aqua><b>MSPT</b></gradient>: <mspt>  -  " +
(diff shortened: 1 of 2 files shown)


canvas-server/src/main/java/io/canvasmc/canvas/RegionizedTpsBar.java
```diff
@@ -70,22 +76,9 @@ public void tick() {
...
-                    .append(Component.text(": ", NamedTextColor.WHITE))
-                    .append(Component.text(msptTruncated, this.worldData.regionData.getRegionSchedulingHandle().ticksToSprint > 0 ? CommandUtil.SPRINTING_COLOR : CommandUtil.getColourForMSPT(mspt)))
-                    .append(Component.text("  -  ", NamedTextColor.WHITE))
-                    .append(gradient("Util", (builder) -> builder.decorate(TextDecoration.BOLD), NamedTextColor.BLUE, NamedTextColor.AQUA))
-                    .append(Component.text(": ", NamedTextColor.WHITE))
-                    .append(Component.text(utilizationTruncated, this.worldData.regionData.getRegionSchedulingHandle().ticksToSprint > 0 ? CommandUtil.SPRINTING_COLOR : CommandUtil.getUtilisationColourRegion(util / 100)));
+            final int players = this.worldData.getPlayerCount();
+            final boolean sprinting = this.worldData.regionData.getRegionSchedulingHandle().ticksToSprint > 0;
+            final Component textComponent = buildComponent(tps, mspt, util, players, sprinting);
```
BaconCat1, line 86: It should catch parse failures and fall back to default, instead of just throwing during ticking.

BaconCat1, line 86: buildComponent() re-reads Config.INSTANCE.tpsBarFormat, performs four String.replace(...) calls, constructs a fresh TagResolver, and reparses MiniMessage every update tick. As that's a lot of churn, is it possible to maybe cache the normalized format?

BaconCat1 (changes requested):

Dueris:
Out of curiosity and since Im not able to check personally ATM, do the gradients match the original tpsbar?

canvas-server/src/main/java/io/canvasmc/canvas/RegionizedTpsBar.java
```diff
@@ -95,6 +90,46 @@ public void tick() {
...
 
+    private Component buildComponent(final double tps, final double mspt, final double utilPercent, final int players, final boolean sprinting) {
+        String raw = Config.INSTANCE.tpsBarFormat;
+        if (raw == null || raw.isBlank()) {
+            raw = DEFAULT_FORMAT;
+        }
+        if (!raw.equals(cachedRawFormat)) {
+            cachedRawFormat = raw;
+            cachedNormalizedFormat = normalize(raw);
+        }
```
BaconCat1: Raw and normalized formats are published independently. This can lead to a mixed state (new raw, old normalized). Consider storing both in a single immutable holder and updating it atomically. While this is likely harmless, it is still better to store them together and update atomically.

BaconCat1 (changes requested):

canvas-server/src/main/java/io/canvasmc/canvas/RegionizedTpsBar.java
```diff
@@ -95,6 +90,46 @@ public void tick() {
...
+
+        TagResolver resolver = TagResolver.builder()
+            .resolver(Placeholder.component("tps", number(tps, TPS_FORMAT, tpsColor)))
+            .resolver(Placeholder.component("mspt", number(mspt, MSPT_FORMAT, msptColor)))
+            .resolver(Placeholder.component("util", number(utilPercent, UTIL_FORMAT, utilColor)))
+            .resolver(Placeholder.component("players", number(players, INT_FORMAT, NamedTextColor.WHITE)))
+            .build();
+
+        try {
+            return MINI_MESSAGE.deserialize(cachedNormalizedFormat, resolver);
```
BaconCat1: The old implementation built the component directly; this version still normalizes only on config changes, but it calls MiniMessage.deserialize(...) and rebuilds the resolver every time the TPS bar refreshes. On Canvas that means a full MiniMessage parse once per second for every active region. Maybe parse/cache this on format changes and keep the tick path to just filling in current values.

BaconCat1:
Besides the above 2, LGTM.

MioArchive:
Hi, thanks for the reviews; I’ll take some time tomorrow to go through everything :)

MioArchive:
Hi, I've fixed both "issues";
- Both fields are now stored together in a single `FormatEntry` record, updated via `AtomicReference`
- On format change, the string is split into pre-parsed static segments and placeholder slots. Each tick only fills in the current values.

Let me know if there is anything else you'd like me to take care of.

Dueris:
Just had a few minor nitpicks I wanted to clean up(annotations). Testing this personally right now, but from overall code-only review, this is looking fine to me

Dueris:
Was there any particular reason why the gradient colors were changed?(Or at least the default)

canvas-server/src/main/java/io/canvasmc/canvas/Config.java
```diff
@@ -485,6 +485,16 @@ public static class Fixes {
     })
     public boolean enableTpsBar = true;
 
+    @Comment({
+        "MiniMessage-formatted line for the TPS bar. Placeholders: <tps>, <mspt>, <util>, <players>.",
+        "Legacy tokens %tps%, %mspt%, %util%, %players% are also accepted and auto-converted."
+    })
+    public String tpsBarFormat =
```
BaconCat1:
```suggestion
    public String tpsBarFormat =
    "<gradient:blue:aqua><b>TPS</b></gradient>: <tps>  -  " +
    "<gradient:blue:aqua><b>MSPT</b></gradient>: <mspt>  -  " +
    "<gradient:blue:aqua><b>Util</b></gradient>: <util>%  -  " +
    "<gradient:blue:aqua><b>Players</b></gradient>: <players>";
```

BaconCat1: This should be the same as the old one. Not sure why that was changed.

MioArchive: Sorry about that; I was just messing around with these values before I even opened the PR, but it should be fixed now : )

MioArchive:
I’ve had another look at everything myself and it seems fine to me for merging; let me know if it isn’t so I can make the necessary changes.

Dueris (approved):
LGTM

Dueris:
Doing some minor cleanup of some tiny things I noticed, then Ill merge

Dueris:
Thank you very much for your contribution, please feel free to contribute again anytime :)

PR #180, opened 2026-03-30
Region-Threading `tick` Command

Dueris:
Canvas previously added back the `tick` command, allowing server-wide modifiable tick rate, sprinting, freezing, etc. This patch was incomplete, because "stepping" was never implemented, and this way was always kinda janky and messy, and a bit all over the place.

This new implementation provides a *replacement* set of logic for the `tick` command, offering the ability to change the server-wide tick rate, and allowing to toggle schedule-handle-specific features like:
- Sprinting -- allows ignoring "deadlines" and speeding up to the next tick immediately until a certain amount of ticks have been run
- Pause -- allows "freezing" the schedule handle, making it so that it will not run specific game elements
- Play -- allows "unfreezing" the schedule handle, now letting it run game elements again
- Walking -- tells the schedule handle to stop sprinting if the schedule handle specified is sprinting

These features allow server admins to modify region-independent tick states, without effecting the rest of the server. There ofc is the option to enable effecting the whole server, since the new command accepts a few arguments:
- `global` - Global tick thread
- `server` - Every single schedule handle
- `<block X coord> <block Z coord>` - To select a specific region. This region **must** be loaded though, and this is only available to players.

This is a much more clean and efficient way to modify and track the independent tick states of regions and the global tick, which can be very useful for debugging, even in production.


Dueris:
Ready for review now, all features are in and tested

Toffikk (approved):

Dueris:
Now that I think abt it, redstone tick time should pause in the region if the region is paused. Ill work on that in the morning probably

Dueris:
image

Redstone tick time is already covered by the if statement here. `runsNormally` is `doesRunGameElements()`

canvas-server/src/main/java/io/canvasmc/canvas/command/sub/RegionTickCommand.java
```diff
@@ -0,0 +1,233 @@
...
+                throw MUST_BE_PLAYER.create();
+            }
+            int chunkX = (first.equalsIgnoreCase("~") ? player.blockPosition.getX() : Integer.parseInt(first)) >> 4;
+            int chunkZ = (second.equalsIgnoreCase("~") ? player.blockPosition.getZ() : Integer.parseInt(second)) >> 4;
+            ThreadedRegionizer.ThreadedRegion<TickRegions.TickRegionData, TickRegions.TickRegionSectionData> region =
+                player.level().regioniser.getRegionAtUnsynchronised(chunkX, chunkZ);
+            if (region == null) {
+                throw NO_REGION_EXISTS.create();
+            }
+            action.accept(TickRegionScheduler.getCurrentRegion().getData().tickHandle);
```
R00tB33rMan: Convert this to `action.accept(region.getData().tickHandle);` so the action is posted to the region at the specified coordinates rather than the command sender's current region

Dueris: Nice catch, thanks

R00tB33rMan (approved):

PR #181, opened 2026-03-30
Packet overflow config options

Dueris:
The changes present in this PR are intended to patch an exploit in regards to excessive opening/closing of containers with lots of NBT data in them causing CPU usage spikes and ping spikes.

This has been realized to be caused by Papers patch for excessive packet sizes for containers. This allows configuring the max packet size for all packets, disabling the excessive packet size fix, and customizing the disconnect reason sent to the client.

The config options are:
- `maximumPacketBytes`- Default to match Vanilla
- `disablePaperPacketOverflowContainerFix` - Keeps the overflow enabled by default
- `packetTooLargeDisconnectReason`

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/network/Connection.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/network/Connection.java.patch
@@ -1,5 +1,14 @@
 --- a/net/minecraft/network/Connection.java
 +++ b/net/minecraft/network/Connection.java
+@@ -213,7 +_,7 @@
+                     if (player != null) player.quitReason = org.bukkit.event.player.PlayerQuitEvent.QuitReason.TIMED_OUT; // Paper - Add API for quit reason
+                     this.disconnect(Component.translatable("disconnect.timeout"));
+                 } else {
+-                    Component component = Component.translatable("disconnect.genericReason", "Internal Exception: " + exception);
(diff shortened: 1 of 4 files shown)


Dueris:
This PR is/will be tested by JackpotMC later today

R00tB33rMan (approved):

R00tB33rMan:
Obviously, plans to move over to Paper in the future, but for now, given this issue's urgency, everything is perfectly fine!

coredex-source (approved):
Looks good to me!

Toffikk (approved):
lgtm. should be reported to paper asap too on their discord exploit report channel

Dueris:
Yea I agree it probably maybe should, but this apparently only happens on Folia. This may not even be the proper fix, but this is what I found that could fix or at least have some sort of layer to protect against these sorts of things

The people doing the exploit thing from what I was told was DuperTroopers guys, and that by spamming open and close a chest on a Folia server filled with shulkers filled with items with lots of NBT data it causes excessive CPU usage and increased ping globally. This was sort of a way to try and find when this sort of thing happens by allowing a toggle of the overflow fallback from Paper(instead of just forcing it enabled) and configuring the max packet bytes.

Im more in favor of just making a PR to Paper with these 3 configs added, since its probably a better idea to do that. Why it only happens on Folia, Im not 100% sure yet

devnootnoot:
We've tested the options thoroughly and seems to work well. We're still monitoring it in production, but so far so good

PR #183, opened 2026-03-31
Add `canvas-supported` Flag

Dueris:
This is for plugins that support Paper and it's forks, excluding Folia, but support Canvas. This acts identical to the folia-supported flag.

This is primarily for plugins who have a setup where they don't support Folia, but support Canvas because of Canvas-only API

This is generally for plugins who can't support Folia, but can support Canvas because of the API Canvas fixes/provides for region threading. Plugins can use this flag or the `folia-supported` flag, as on Canvas, they both do the same thing, but in Folia if you only have `canvas-supported` then it won't load it because Folia needs `folia-supported`, but Canvas can load plugins without `canvas-supported`(requires `folia-supported` then)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -108,6 +108,10 @@ subprojects {
             authors = listOf("CanvasMC")
             foliaSupported = true
         }
+
+        tasks.processResources {
+            duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -108,6 +108,10 @@ subprojects {
             authors = listOf("CanvasMC")
             foliaSupported = true
         }
+
```
Toffikk, line 111: why this change? otherwise lgtm

Dueris: I was using one of my local test plugins to test the functionality, however with the paper plugin generator gradle plugin, it was causing duplicate handling strategy errors, so I made it exclude so that it would just take the one in src/main/resources if present, and if it isn't present then it auto-gens

MrMasrozYTLIVE (approved):
Lgtm

Toffikk (approved):
a long needed change xD

R00tB33rMan (approved):
LGTM - excited to see plugins utilize Canvas more intuitively!

PR #186, opened 2026-04-06
Replace respawn event and unload event

Dueris:
To follow along with our teleport events and portal events, we've made the decision to also replace the world unload event and the player teleport event.

Restoring these events realistically speaking can cause instabilities with plugins compatible with Folia, since Folia doesn't execute these events. If plugins utilize the 2 events that are now replaced in Paper, but supports Folia too, then this could cause instability issues in the plugin, causing crashes.

By replacing these with our own API, we ensure that Folia-compatible plugins remain compatible with Canvas, and we introduce API that allows plugin developers to hook into these new events


R00tB33rMan (approved):
LGTM - looks like we have everything covered here from an outside POV

PR #190, opened 2026-04-11
Optimize Netty flushing in region threading

Dueris:
In Vanilla, it suspends/resumes flushing of packets unless the packet being sent is async. This helps have fewer flush calls meaning less syscalls and such, and is much better for TCP connections and Netty

Folia is always in a constant state of flush, meaning every single packet makes the server flush the channel. This is ungodly stupid and is actually horrendous for performance of Netty.

This patch fixes this and makes it restore the suspend/resume flushing during the main region tick, and optimizes the entity thread check to call `Thread.currentThread()` fewer times.

This is moved to the region threading fixes patch, since technically this is a fix to the region threading model, however it can be moved outside of this into its own patch.

### DIFF
--- a/build-data/canvas.at
+++ b/build-data/canvas.at
@@ -28,6 +28,10 @@ public io.papermc.paper.threadedregions.TickRegionScheduler$RegionScheduleHandle
 public io.papermc.paper.threadedregions.TickRegionScheduler$RegionScheduleHandle tickTimes1m
 public io.papermc.paper.threadedregions.TickRegionScheduler$RegionScheduleHandle tickTimes5m
 public io.papermc.paper.threadedregions.TickRegionScheduler$RegionScheduleHandle tickTimes5s
+public io.papermc.paper.threadedregions.TickRegionScheduler.TickThreadRunner
+public io.papermc.paper.threadedregions.TickRegionScheduler.TickThreadRunner currentTickingRegion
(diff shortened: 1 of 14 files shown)


Dueris:
This is being tested in production and has been tested for the last 3ish hours already, no issues have arisen.

Dueris:
The recent changes I just made, commit [00da867](https://github.com/CraftCanvasMC/Canvas/pull/190/commits/00da8571652bbcbf24a9b749c7ac07ea70a7d7b7) to be exact, makes it so it reduces flush calls even further, to only sending a flush if a non-flushed packet is sent. This way, it doesnt always flush 1 time every tick for every player in the region. Now, it will only flush on resume if a player has a non-flushed packet that was sent while being suspended.

This helps optimize situations where no packets were sent to the client during the tick, or all the packets sent to the client were off the owning thread of the player, meaning they were sent already. This makes it so it only flushes when it has to

Dueris:
Also, should documentation be added to the website?

devnootnoot (approved):
Looks good!

devnootnoot:
> Also, should documentation be added to the website?

Probably, yes, would briefly mention it somewhere.

R00tB33rMan (approved):

R00tB33rMan (approved):

PR #191, opened 2026-04-11
Fix: Subspace Bubble advancement bug fix with respawning

devnootnoot:
On folia, respawning while in the Nether would grant you the Subspace Bubble advancement, when it's not supposed to.

This patch fixes that by just simply passing whether the player is respawning, and only counting the distance travelled if they are not respawning.

This has been tested locally, and works fine.


canvas-server/minecraft-patches/base/0022-Subspace-advancement-respawning-fix.patch
```diff
@@ -0,0 +1,130 @@
...
++                                      EntityTreeNode passengerTree, java.util.function.Consumer<Entity> teleportComplete, boolean respawning) { // Canvas - region threading
+         Vec3 pos = this.position();
+         ChunkPos posChunk = new ChunkPos(
+             ca.spottedleaf.moonrise.common.util.CoordinateUtils.getChunkX(pos),
+@@ -4224,7 +4230,7 @@ public abstract class Entity implements SyncedDataHolder, DebugValueSource, Name
+                     );
+                     List<EntityTreeNode> fullTree = passengerTree.getFullTree();
+                     for (EntityTreeNode node : fullTree) {
+-                        node.root.placeSingleSync(originWorld, destination, node, teleportFlags);
++                        node.root.placeSingleSync(originWorld, destination, node, teleportFlags, respawning); // Canvas - region threading
```
Dueris, line 113: Unsure if we really need to pass this, since any passengers of a dying/respawning entity would probably be unmounted on death, right? So would this rly matter if we pass `respawning`?

Dueris (changes requested):

canvas-server/minecraft-patches/base/0022-Subspace-advancement-respawning-fix.patch
```diff
@@ -0,0 +1,130 @@
...
+      */
+     public static final long TELEPORT_FLAG_UNMOUNT = 1L << 2;
+ 
+-    protected void placeSingleSync(ServerLevel originWorld, ServerLevel destination, EntityTreeNode treeNode, long teleportFlags) {
++    protected void placeSingleSync(ServerLevel originWorld, ServerLevel destination, EntityTreeNode treeNode, long teleportFlags, boolean respawning) { // Canvas - region threading
+         destination.addDuringTeleport(this);
+     }
++    // Canvas start - region threading
++
++    protected void placeSingleSync(ServerLevel originWorld, ServerLevel destination, EntityTreeNode treeNode, long teleportFlags) {
```
Dueris, line 97: Circular call, constantly loops recursively until stack overflow

Dueris (changes requested):

canvas-server/minecraft-patches/base/0022-Subspace-advancement-respawning-fix.patch
devnootnoot: True, didn't notice, will fix. Thanks.

canvas-server/minecraft-patches/base/0022-Subspace-advancement-respawning-fix.patch
devnootnoot: It's not really good to assume that, the entity dying (not just the passenger) is part of the full passenger tree, so unless we start separating the parent vs passengers, I think we do need to pass it along

Dueris: But do dying players take their passengers with them on respawn?

Dueris: And by this point in the code, the player is already removed from the world

devnootnoot: yes, but the `placeSingleSync()` method is triggered when the actual player respawns as well, and that triggers advancement checks, which needs to know if the player was respawning or not (it's actually only triggered in there for respawns i'm pretty sure)

devnootnoot: > And by this point in the code, the player is already removed from the world

the respawning state is still figured out before they do though, so not really sure why this matters?

canvas-server/minecraft-patches/base/0022-Subspace-advancement-respawning-fix.patch
devnootnoot: Figured out the fix for this, will be in the next commit.

canvas-server/minecraft-patches/base/0022-Subspace-advancement-respawning-fix.patch
Dueris: AH I see, mb I overlooked that. I was just lookin at the diff and noticed that, I forgot that the "passenger tree" also includes the root entity in the passenger tree off the top of my head, mb

devnootnoot (commented):
Please check if recursion issue is fixed now @Dueris

R00tB33rMan (approved):

Dueris (approved):

PR #192, opened 2026-04-14
Aggressive Thread Checks

Dueris:
Notes:

- name-related methods access synched entity data, not at all safe to access
- pointer fetching is safe, the pointer resolvers need thread checked methods
- permission based API methods in CraftEntity use a shared instance, doesn't need to be thread checked

Don't merge yet, WIP

`org.bukkit.craftbukkit.entity.boats` has no overriding methods, no need for thread checks


Dueris:
Do note, the permissionable API is not thread-safe, but should be made thread safe. This is a shared instance across all entities, as such it should be thread-safe

Dueris:
Do note, realistically speaking we dont need to thread guard methods calling `getHandle` unless it's `CraftEntity`, `CraftLivingEntity`, `CraftHumanEntity`, or `CraftPlayer`. This is because `getHandle` has a thread guard built in from Folia for all of these classes, but the `CraftPlayer#getHandle` method doesn't have this tick guard intentionally, so we need to guard every method in the player and its super classes to ensure all methods are thread checked appropriately

Dueris:
Should we thread-check the playerlist header/footer getters and setters in `CraftPlayer`?

Dueris:
Also we should check `CraftPlayer#pluginWeakReferences` field, it's not thread-safe and is a static field used in the visibility API

Dueris:
Folia disables the player thread checks, for good reason. After lots of testing, I have realized that this PR actually overdoes these checks, causing even internals to start tripping tick guards.

It should be decided eventually whether or not it is worth it to keep this PR and eventually merge it, or if it should just be fixed

Dueris:
Requires rebase @CraftCanvasMC/core-team

Toffikk:
> Requires rebase @CraftCanvasMC/core-team

done

Dueris:
Me and RootBeer will call soon to go over this PR thoroughly. All the initial changes are done tho and ready to be tested/reviewed.

Dueris:
This is being closed due to design flaws. An iteration of this may follow, but for now this is way too sensitive to plugins

PR #208, opened 2026-04-28
Fix filtered ender pearl motion sync

ssquadteam:
Fixes ender pearls being affected by `filterClientboundSetEntityMotionPacket`.

Ender pearls inherit shooter movement as part of their velocity. When motion packets are filtered, the client can miss velocity corrections and visually desync pearl movement, especially when throwing while sprinting.

This adds `ThrownEnderpearl` to the existing entity motion packet allow-list, matching other entities that require velocity sync for smooth client-side rendering.

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/server/level/ServerEntity.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/server/level/ServerEntity.java.patch
@@ -58,7 +58,7 @@
                      Vec3 deltaMovement = this.entity.getDeltaMovement();
                      double d = deltaMovement.distanceToSqr(this.lastSentMovement);
                      if (d > 1.0E-7 || d > 0.0 && deltaMovement.lengthSqr() == 0.0) {
-@@ -210,9 +_,41 @@
+@@ -210,9 +_,42 @@
                                          )
                                      )
                                  );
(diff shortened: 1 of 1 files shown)


Toffikk:
If you could rebase this for the dev/26.1.2 branch, that’d be awesome

ssquadteam:
> If you could rebase this for the dev/26.1.2 branch, that’d be awesome

Will do !

canvas-server/minecraft-patches/sources/net/minecraft/server/level/ServerEntity.java.patch
```diff
@@ -85,14 +85,14 @@
 +                               3. Ender Eyes, they depend on their velocity being set bc the packet sets the velocity on the client, but the client doesn't handle the
 +                                    ender eye velocity on its own bc the ender eye velocity is in relation to the stronghold structure which the client doesn't have access to
 +                               4. Shulker bullets use random to determine where they are going, so we need to send velocity packets to ensure it remains smooth on the client
-+
```
Toffikk, line 88: add a comment explaining why ender pearls are exempt

Toffikk (changes requested):

Dueris:
I wanted to try and help out with this since I do see the issue with this, however this is noticeable(at least from my view) both standing still and while flying in this case, I do see the SLIGHTEST bit of choppiness, however I can replicate even with the velocity packet filter being disabled

Dueris:
I actually think Im gonna close this, since while I do think this is kinda a wonky issue, I dont actually think this is Canvas. I was able to sorta replicate this on Folia, perhaps something in Canvas makes this worse, but Im not quite sure what to do about this yet, Ill investigate this more later, but I think this falls under the category of "not entirely a Canvas issue I think"

PR #210, opened 2026-04-30
Fix save task player data

Euphillya:
I noticed on a server I was working on that players were being rolled back if they didn’t disconnect properly (e.g., machine crash or similar). After investigating, I found that PlayerList#saveAll was never triggered because Player#lastSave was set to Long.MIN_VALUE when the player connected. Since this is a negative value, the condition required to perform the save was never met.

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/server/players/PlayerList.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/server/players/PlayerList.java.patch
@@ -35,10 +35,11 @@
              )
          );
          player.getBukkitEntity().sendSupportedChannels(); // CraftBukkit
-@@ -292,6 +_,7 @@
+@@ -292,6 +_,8 @@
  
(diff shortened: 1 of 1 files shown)


Dueris (approved):
Could this be rebased to 26.1 though? No joke, less than an hour ago we ended production of 1.21.11 builds in prep for 26.1 releasing tmrw

Dueris:
Also has this been tested in production?

Euphillya:
> Also has this been tested in production?

I will rebase on 26.1

> Also has this been tested in production?

Yes but the server I was working on wasn't running on a fork of Canvas but on Folia, and I reproduced the same problem with Canvas (probably a Folia bug to begin with).

Dueris:
Alrighty. I did check Folias source before reviewing and saw that the issue was indeed Folia. Im happy to merge this once this is rebased.

Nice work

Euphillya:
Good, I changed the branch and rebased it.

Toffikk (approved):

Toffikk:
@Euphillya It might be worth to contribute this to upstream too

PR #212, opened 2026-05-06
Region Guards

Dueris:
It is seen time and time again that plugins are doing blatantly unsafe practices off the owning context of certain entities and players. With players, this sort of abuse is easy, since they are not thread-checked like non-player-entities are. There ofc is no real way to introduce this, since this logic is intentional, and by thread checking practically everything in the Player Bukkit API, plugins will not be compatible in the slightest with Canvas.

The proposal here is to implement "guards", which implement thread checks with optional severity of the check, on places that are commonly seen in stacktraces and crashes of issues caused by plugins, like inventory opening/closing for example.

There are 3 options for this, `SILENT`, `LOG`, `THROW`. `SILENT` means nothing is logged or checked, `LOG` means the server logs a warning, and `THROW` acts as a throwing thread check.

In this PR there are also a few hard-thread-checks, since there are methods that simply should never be allowed to just be called async and the server owner turns off the warning. Those 2 are just save/load player data on the `Player` object.

The hope with this is to narrow down plugin issues, and reduce crashes induced by said plugins unsafely scheduling operations. If there are more things that should be guarded, or suggestions, or questions, comment here.

### DIFF
--- a/canvas-server/src/main/java/io/canvasmc/canvas/util/TickGuard.java
+++ b/canvas-server/src/main/java/io/canvasmc/canvas/util/TickGuard.java
@@ -1,0 +1,51 @@
+package io.canvasmc.canvas.util;
+
+import ca.spottedleaf.moonrise.common.util.EntityUtil;
+import ca.spottedleaf.moonrise.common.util.TickThread;
+import ca.spottedleaf.moonrise.common.util.WorldUtil;
+import io.canvasmc.canvas.GlobalConfiguration;
+import net.minecraft.core.BlockPos;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.level.Level;
+
+import static ca.spottedleaf.moonrise.common.util.TickThread.getThreadContext;
+import static io.canvasmc.canvas.GlobalConfiguration.LOGGER;
+
+public class TickGuard {
+
+    public static void guard(final BlockPos pos, final Level world, String reason) {
+        switch (GlobalConfiguration.getInstance().regionScheduler.guardSeverity) {
+            case SILENT -> ensureIsTickThread(reason);
+            case LOG -> {
+                // ensure tick thread first, since that is required, then we check and log
+                ensureIsTickThread(reason);
+                if (!TickThread.isTickThreadFor(world, pos)) {
+                    LOGGER.warn("Thread failed main thread check: {}, context={}, world={}, block_pos={}", reason, getThreadContext(), WorldUtil.getWorldName(world), pos, new Throwable());
+                }
+            }
+            case THROW -> TickThread.ensureTickThread(world, pos, reason);
+        }
+    }
+
+    public static void guard(final Entity entity, String reason) {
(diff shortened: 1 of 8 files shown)


R00tB33rMan (commented):
We should likely include any form of attribute modification in this region tick guard mix, since attributes generally throw silently and are the cause of a good number of hard-to-catch crashes.

coredex-source (approved):

coredex-source:
Other than Root's request everything looks fine.

Dueris:
I mean, attributes are a bit hard to do given the AttributeMap doesn't store the owning entity reference IIRC. It's maybe possible to store it though, Ill take another look later

Dueris:
Is it worth it to just hard check these, instead of making them configurable? While it may be useful to be configurable, the checks introduced here should probably be forced, since they do cause race conditions and cause issues if run improperly, and it's generally safer to just throw an error and fix it in the plugin than just log a warning that will just be ignored by owners. While a crash isn't favorable for this sort of thing, isn't it kinda what's necessary to fix these issues or at least get them noticed?

040Elias:
I think the configurable approach is worth keeping. The concern about logs being silently ignored is valid — but the fix is to make THROW the default, not to remove configurability. If owners are stuck with a broken third-party plugin, they need a way to temporarily downgrade to LOG. Hard-checking everything will just push them to a fork instead of pressuring plugin devs to actually fix their code.

Dueris:
Fair enough, Ill set the default to `THROW` later today

Toffikk (approved):

canvas-server/src/main/java/io/canvasmc/canvas/util/TickGuard.java
```diff
@@ -0,0 +1,56 @@
...
+        switch (GlobalConfiguration.getInstance().regionScheduler.guardSeverity) {
+            case SILENT -> ensureIsTickThread(reason);
+            case LOG -> {
+                // ensure tick thread first, since that is required, then we check and log
+                ensureIsTickThread(reason);
+                if (!TickThread.isTickThreadFor(entity)) {
+                    LOGGER.warn("Thread failed main thread check: {}, context={}, entity={}", reason, getThreadContext(), EntityUtil.dumpEntity(entity), new Throwable());
+                }
+            }
+            case THROW -> TickThread.ensureTickThread(entity, reason);
```
MrMasrozYTLIVE, line 46: How is that different from SILENT with ensureIsTickThread(reason) method? (My brain is not braining at 5AM)

MrMasrozYTLIVE (approved):
LGTM

canvas-server/src/main/java/io/canvasmc/canvas/util/TickGuard.java
Dueris: Because we replace certain calls to the async catcher with the region gaurds, which means if we do silent as literally do nothing, we end up removing thread checks that were there before for *generic* tick threads. Silent just runs a a check for if the current thread `instanceof` a `TickThread`, while the other checks check that alongside the region context

R00tB33rMan (approved):

PR #215, opened 2026-05-06
perf: O(1) runner lookup + cached overload threshold in affinity scheduler

040Elias:
## Summary

Two targeted micro-optimizations in the affinity scheduler hot path.

### 1. O(1) `getCurrentTickThreadRunner()` via `ThreadLocal`

**Before:** `AffinitySchedulerThreadPool#getCurrentTickThreadRunner()` did a linear scan over `runners[]`, comparing each runner's backing thread to `Thread.currentThread()`. Cost was O(thread_count) per call.

**After:** Each `TickThreadRunner` registers itself into a `ThreadLocal<TickThreadRunner>` at the top of `run()`. `getCurrentTickThreadRunner()` reads this ThreadLocal — O(1), no allocation, no lock.

Called from `AffinityHandler.tryTransferPinningState()` during every region merge, split, and profiler-initiated handle transfer. On a 16-thread server this was 16 reference comparisons per event; now it's a single ThreadLocal read.

### 2. Cached overload threshold in `ScheduledHandleTickState`

**Before:** `tickStart()` — called every tick on every region — evaluated:
```java
GlobalConfiguration.getInstance().regionScheduler.overloadedLogMillis * 1_000_000L
```
on each invocation. Three levels of indirection + a 64-bit multiply, every tick.

**After:** `overloadedLogNanos` is computed once in the constructor and stored as a `final long`. The config value does not change at runtime.

## Changes

- `AffinitySchedulerThreadPool`: add `ThreadLocal<TickThreadRunner> runnerLocal`, populate it in `TickThreadRunner.run()`, use it in `getCurrentTickThreadRunner()`
- - `ScheduledHandleTickState`: add `final long overloadedLogNanos` field, initialize in constructor, replace inline expression in `tickStart()`
## Performance Impact

- **Runner lookup:** eliminates O(n) array scan; becomes a single `ThreadLocal.get()` (~5 ns). At 16 threads the old path touched ~16 cache lines; ThreadLocal reads from thread-local storage with no contention.
- - **Overload check:** eliminates a static method call, two field dereferences, and a multiply every tick. With 100 regions ticking at 20 TPS that's 2,000 unnecessary multiplications/second eliminated.
## Testing

- Start a Canvas server with the AFFINITY scheduler configured
- - Spawn enough players/chunks to create multiple regions
- - Enable region profiling and trigger a region merge/split — verify profiler transfer still works correctly
- - Confirm overload warnings still fire when a region misses its deadline
## Risks

- **ThreadLocal lifetime:** threads are never reused in this pool so the ThreadLocal is naturally cleaned up with the thread. No leak risk.
- - **Config hot-reload:** `overloadedLogNanos` is computed at handle construction. If Canvas ever adds runtime config reload the cached value would be stale — the field name makes the intent explicit and easy to revisit.
- - No behavioral changes; purely structural/caching improvement.

### DIFF
--- a/canvas-server/src/main/java/io/canvasmc/canvas/GlobalConfiguration.java
+++ b/canvas-server/src/main/java/io/canvasmc/canvas/GlobalConfiguration.java
@@ -148,6 +148,15 @@ public class GlobalConfiguration extends Part {
                 // update all info with player, covers 1.8 combat config
                 playerList.sendAllPlayerInfo(player);
             }
+
+            // Canvas start - push updated overload threshold to all live ScheduledHandleTickState instances
+            io.papermc.paper.threadedregions.RegionizedServer.getGlobalTickData().getTickManager().onConfigReload();
+            for (final net.minecraft.server.level.ServerLevel world : MinecraftServer.getServer().getAllLevels()) {
+                world.regioniser.computeForAllRegionsUnsynchronised(region ->
+                    region.getData().getRegionSchedulingHandle().getTickManager().onConfigReload()
+                );
+            }
+            // Canvas end
         }
         else {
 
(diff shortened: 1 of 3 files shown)


Dueris:
I don't entirely see the necessity to cache the config lookup and add a "on config reload" for that

Also I don't like the idea of a threadlocal, but I do agree that a linear scan maybe isnt the best approach

PR #217, opened 2026-05-08
Villager POI Search Range Configs

Dueris:
AcquirePoi searches always run with a range of `48` blocks. Villagers have 4 POI types that are run through 3 different behavior controllers in the mob brain, specifically searching for `HOME`, `MEETING`, `JOB_SITE`, and `POTENTIAL_JOB_SITE`. Note that `JOB_SITE` is the memory to validate and `POTENTIAL_JOB_SITE` is the memory to acquire in their shared behavior controller. These POIs are later validated once present via the `ValidateNearbyPoi` class, which actually validates them in a radius of `16` blocks, unlike `48` which is what the initial search is for.

With this in mind, there are 3 configs added that shrink the search radius from `48` to `16`, to match the validator. These configs are as such:
- `reduceJobSitePoiSearchRange` - Reduces the `JOB_SITE`/`POTENTIAL_JOB_SITE` POI search range
- `reduceHomePoiSearchRange` - Reduces the `HOME` POI search range
- `reduceMeetingPointPoiSearchRange` - Reduces the `MEETING_POINT` POI search range

Note that these configs are in kebab-case for the actual YAML config

This PR also disables these 3 behavior controllers from loading POI chunks if the chunk isn't currently loaded. There were some noticeable MSPT improvements, being roughly `18.21ms` -> `13.57ms` of Villager ticking(inactive tick included), in a single region with ~585 villagers. While not the most impactful optimization, this does help optimize the current heaviest behavior controllers for Villagers in Canvas, which is at least *something*.

Note that these options do cause slight deviation from Vanilla, however probably not entirely noticeable to the average player.

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/ai/behavior/AcquirePoi.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/ai/behavior/AcquirePoi.java.patch
@@ -1,0 +1,33 @@
+--- a/net/minecraft/world/entity/ai/behavior/AcquirePoi.java
++++ b/net/minecraft/world/entity/ai/behavior/AcquirePoi.java
+@@ -54,6 +_,21 @@
+         final Optional<Byte> onPoiAcquisitionEvent,
+         final BiPredicate<ServerLevel, BlockPos> validPoi
+     ) {
++    // Canvas start - configurable scan range and if poi should load
++        return create(poiType, memoryToValidate, memoryToAcquire, onlyIfAdult, onPoiAcquisitionEvent, validPoi, (_) -> SCAN_RANGE, ca.spottedleaf.moonrise.patches.poi_lookup.PoiAccess.LOAD_FOR_SEARCHING);
++    }
++
++    public static BehaviorControl<PathfinderMob> create(
++        final Predicate<Holder<PoiType>> poiType,
++        final MemoryModuleType<GlobalPos> memoryToValidate,
++        final MemoryModuleType<GlobalPos> memoryToAcquire,
++        final boolean onlyIfAdult,
++        final Optional<Byte> onPoiAcquisitionEvent,
++        final BiPredicate<ServerLevel, BlockPos> validPoi,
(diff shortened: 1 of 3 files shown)


MrMasrozYTLIVE (approved):
Lgtm

coredex-source (approved):

R00tB33rMan (approved):
Don’t see any breakages this’d ensue as-is and looks like all sites are covered!

devnootnoot:
I like the idea behind this, but I think this needs one change before merging.

Right now the range configs only control whether the search range is 16 or 48. However, the behavior controllers always pass `false` for `shouldLoad`, even when the new configs are disabled. That means the PR changes vanilla/Paper behavior even with all 3 options set to false, because villagers will no longer load chunks during these AcquirePoi searches, right?

So if a user leaves:

`reduce-job-site-poi-search-range: false`
`reduce-home-poi-search-range: false`
`reduce-meeting-point-poi-search-range: false`

they would probably expect vanilla-like behavior with a 48 block search range, but the chunk loading behavior is still changed.

I think the chunk-loading change should either have its own config, or `shouldLoad` should stay as `PoiAccess.LOAD_FOR_SEARCHING` when the relevant optimization is disabled. 

I am also not fully convinced that matching AcquirePoi’s range to ValidateNearbyPoi’s 16 block radius is logically equivalent. Acquisition and validation are different steps. A villager may need to discover a bed/workstation/bell farther than 16 blocks away, while validation later only checks nearby/known POIs.

Could you split or adjust this so:
- disabling the new configs preserves both the 48 block range and the original chunk-loading behavior
- the no-chunk-load behavior is separately configurable, or tied clearly to an enabled optimization
- the docs mention that reducing the range may prevent villagers from acquiring POIs between 17 and 48 blocks away

With that fixed, I think this would be much safer to merge.

Dueris:
Oh yea that makes sense

Ill get on that in a tiny bit

Dueris:
Fixed up

devnootnoot:
Looks good to me now, assuming this has been tested.

devnootnoot (approved):

PR #218, opened 2026-05-09
Fix block entity data loss and hopper sync

RuleGaed:


Dueris:
Ive fixed up this PR rq

Ive changed the Hopper BE patch to just recompute the full state entirely

Ive also removed your patch on `LevelChunk`, since literally 2 lines below the line you added is:
```java
this.pendingBlockEntities.putAll(protoChunk.getBlockEntityNbts());
```
Where `getBlockEntityNbts()` returns an unmodifiable map of `pendingBlockEntities`

Dueris (approved):

PR #219, opened 2026-05-10
Restore spawner configs, implement 2 new configs

Dueris:
This restores the 1.21.11 spawner configs that were dropped in 26.1.2's update, now placing it in the `blocks` section, alongside 2 new configs:

- `chests-can-open-with-full-block-above`
- `full-chiseled-book-shelves-count-as-valid-enchant-power-sources`

These 2 add the capability to open chest blocks while they have a block above, and also use filled chiseled bookshelves as a valid power source for enchantment tables

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/LivingEntity.java.patch
@@ -239,7 +239,7 @@
  
      @Override
      public boolean isCollidable(boolean ignoreClimbing) {
-+        if (this.canvas$fromSpawner && level().canvasConfig().spawner.spawnedEntitiesHaveNoCollision) return false; // Canvas - spawner configs
++        if (this.canvas$fromSpawner && level().canvasConfig().blocks.spawner.spawnedEntitiesHaveNoCollision) return false; // Canvas - spawner configs
          return this.isAlive() && !this.isSpectator() && (ignoreClimbing || !this.onClimbable()) && this.collides; // CraftBukkit
          // Paper end - Climbing should not bypass cramming gamerule
      }
--- a/canvas-server/minecraft-patches/sources/net/minecraft/world/level/BaseSpawner.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/world/level/BaseSpawner.java.patch
@@ -1,5 +1,89 @@
 --- a/net/minecraft/world/level/BaseSpawner.java
 +++ b/net/minecraft/world/level/BaseSpawner.java
+@@ -45,14 +_,77 @@
+     public @Nullable SpawnData nextSpawnData;
+     private double spin;
+     private double oSpin;
+-    public int minSpawnDelay = 200;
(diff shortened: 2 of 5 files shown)


R00tB33rMan (approved):

coredex-source (approved):

devnootnoot:
Looks mostly good, but I have two concerns before merging.

The new block configs themselves seem fine to me, my bigger concern is with the restored spawner configs.

Right now the spawner config values are used as defaults/fallbacks for the `BaseSpawner` fields and for missing NBT values. For example, `MinSpawnDelay`, `MaxSpawnDelay`, `SpawnCount`, `MaxNearbyEntities`, `RequiredPlayerRange`, and `SpawnRange` still prefer the NBT value if present, and only fall back to the Canvas config if the tag is missing.

That means this may not behave like a true world config override for existing spawners. Existing spawners that already have these values saved in NBT will keep using their saved values, so changing the Canvas config later may not affect them. If that is intended, the docs should say these are default values for newly-created / missing-NBT spawners, not global overrides for all spawners. If the intended behavior is to control all spawners globally, this implementation probably needs to apply the config during ticking instead of only during construction/load fallback.

I would also add validation to the new spawner config values:
- `min-spawn-delay >= 0`
- `max-spawn-delay >= min-spawn-delay`
- `spawn-count >= 0`
- `max-nearby-entities >= 0`
- `required-player-range >= 0`
- `spawn-range >= 0`

Without that, a bad config like `min-spawn-delay: 800` and `max-spawn-delay: 200`, or negative ranges/counts, could cause weird behavior or potentially break delay/random logic depending on the upstream implementation.

Small side note: the hopper-related `canvas$getContainerForHoppers` addition in `ChestBlock` looks unrelated to this PR title/description. If it is intentionally part of this PR, I would mention it in the PR description.

Dueris:
It's not possible to make it something that updates with block-entities during ticking, it's only possible for creation and at creation time only and for fallback.
 
This is intentional by design because Paper/Bukkit API allows u to set these at runtime. It's also read from NBT later when loaded back in, which could have completely different values than the defaults. It becomes a whole mess so for simplicity it's kept as just on creation

I do agree tho that it should be noted that it's only at creation, I was thinking of that earlier today when I was going over this PR again. I will add the validators, I forgot about those lol.

Thank you, will get those done later

Dueris:
The method ur talking about tho is just a format fix, not an addition, it was caught by me when rebuilding patches for the new chest config

devnootnoot:
Yeah, if there is no solution for having it update existing values, then just a note somewhere saying that it's only for creation works for me, just make sure it's explained very well so users aren't confused.

My bad about the extra method, I didn't really look through that properly, just saw something that I thought was unrelated and thought I should mention it.

Let me know when the validators are done and I'll re-review it!

devnootnoot:
LGTM now, but would probably be nice if the "This is only applied at spawner creation" was explained a little more in depth, like what happens to existing spawners? I feel like it's pretty important for server owners to know what exactly you mean by "spawner creation".

Other than that, looks great.

devnootnoot (approved):

PR #220, opened 2026-05-10
simd-support for chunk-gen

yunuservices:
## What changed?

  - Added SIMD helpers for chunk generation.
  - Integrated SIMD-assisted paths into NoiseChunk and Aquifer hot paths.
  - Added a new config option: chunk-system.simd-chunk-generation.
  - This setting is startup-only (requires restart).
  - Added AVX3F/AM5 notes and startup logs.
  - Adjusted the AVX3F compatibility threshold for better AM5/JVM behavior.

  ## Benchmark notes

  - On my AM5 benchmarks, I did not observe regressions.
  - I have not tested AM4 yet. Although it should be supported, AM4 may show different behavior depending on SIMD execution characteristics.
  - I observed a notable CPS increase in my test setup:
      - CPU: Ryzen 7 7700X
      - Threads: 6 (server), 5 (Chunky workers)
      - Memory: 10 GiB container
      - Result: ~240 CPS -> ~320 CPS

  ## Testing note

  **Please add --add-modules=jdk.incubator.vector to startup flags (before -jar) when testing this PR.**

### DIFF
--- a/canvas-api/src/main/java/io/canvasmc/canvas/simd/VectorizedChunkGeneration.java
+++ b/canvas-api/src/main/java/io/canvasmc/canvas/simd/VectorizedChunkGeneration.java
@@ -1,0 +1,122 @@
+package io.canvasmc.canvas.simd;
+
+import jdk.incubator.vector.IntVector;
+import jdk.incubator.vector.VectorMask;
+import jdk.incubator.vector.VectorSpecies;
+
+public final class VectorizedChunkGeneration {
+
+    private static final VectorSpecies<Integer> I_SPEC = IntVector.SPECIES_PREFERRED;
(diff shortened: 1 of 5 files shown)


Toffikk:
Hi, thanks for your PR!
I wanted to ask if there are any known deviations from vanilla, because if so, then they will need to be fixed before this can get merged

Toffikk:
Another thing that would help us, would be providing some benchmarks (and maybe a link to them so we can benchmark ourselves too) and spark reports to see the impact of this change.

Toffikk:
Due to a lack of response and a possible use of an LLM, we’ll be closing this PR.
Feel free to open a new one once the changes are verified not to deviate from Vanilla and confirmed to bring benefits

PR #221, opened 2026-05-11
Fix stale player list GUI

RuleGaed:
Fixes a stale player list GUI issue where the list could fail to update if one player leaves and another joins between refreshes. The previous logic skipped updates when the player count stayed the same, but the displayed player names may still have changed.


RuleGaed:
i removed the player list size check because it causes a bug where the gui doesnt update if a player leaves and another joins simultaneously since the list size remains unchanged but the names have actually changed this ensures the list is always accurate

Toffikk (changes requested):
Generally, its preferred if you separate multiple unrelated changes into multiple PRs rather than merge all of them into one

RuleGaed:
i separated the changes and updated this PR to focus only on the player list gui fix i removed the other unrelated optimizations to keep it simple as suggested thanks for the feedback

Dueris (approved):

R00tB33rMan (approved):

Toffikk (approved):

PR #223, opened 2026-05-11
Fix race condition in AffinitySchedulerThreadPool

RuleGaed:
- Replaced the `int nextSteal` variable with `AtomicInteger` to ensure thread-safe work stealing across multiple tick threads.
- Updated the steal logic to use `getAndUpdate` for atomic increments and modulo operations.
- This prevents potential race conditions where threads might skip tasks or conflict during high-load scheduling.


Dueris:
image
I dont know if all of your changes went through, I only see this for the changes you made.

RuleGaed:
my bad i forgot to include the usage update in the previous commit i just forced pushed the fix now you should be able to see the getandupdate logic in the stealing part as well sorry for the confusion

Dueris (approved):

Dueris (approved):

R00tB33rMan (approved):
Very nice catch!

RuleGaed:
Thank you

PR #224, opened 2026-05-11
Optimize block entity map fetching with intelligent caching

RuleGaed:
- Added a caching mechanism for `canvas$buildBlockEntityMap` in `ChunkAccess` to avoid redundant map allocations.
- The cache is automatically invalidated whenever a block entity is added, removed, or all are cleared.
- This significantly improves performance for block-entity-heavy operations (like hoppers) by reusing the unmodifiable map until the underlying data changes.

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/world/level/chunk/ChunkAccess.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/world/level/chunk/ChunkAccess.java.patch
@@ -43,6 +43,8 @@
 +        return bucket[canvas$sectionIndex(globalX, globalZ)];
 +    }
 +
++    private Map<BlockPos, BlockEntity> canvas$cachedBlockEntityMap = null;
++
 +    @Nullable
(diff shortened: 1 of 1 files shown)


Dueris (approved):
LGTM

I will note tho that I'm not sure where in Minecraft the map method is called, I believe when I initially made the changes I changed all access points too to use the optimized implementations

Regardless tho, I do like this, looks fine to me

R00tB33rMan (approved):

Toffikk (approved):

devnootnoot:
Looks good to me as long as `canvas$buildBlockEntityMap()` is never expected to be modifiable. 

Also, the cache is only safe if every modification to `canvas$blockEntitySections` goes through the patched set/remove/clear methods. If anything writes directly into the section arrays without clearing `canvas$cachedBlockEntityMap`, the returned map can become stale.

As long as those two things are checked, then should be fine to merge.

devnootnoot:
After a talk with @RuleGaed in DMs, my concerns have been answered, and everything looks good to me now.

devnootnoot (approved):

PR #225, opened 2026-05-11
Implement customizable configuration mappers

RuleGaed:
Implements a registry-based system for configuration object mappers, allowing dynamic support for types like Identifier and ResourceKey without hardcoding them in the motor


canvas-server/src/main/java/io/canvasmc/canvas/tick/AffinitySchedulerThreadPool.java
Dueris, line 77: Im not quite sure what happened here, but this is already in the 26.1 branch

PR #226, opened 2026-05-11
Add customizable configuration mappers

RuleGaed:
Adds a registry-based system for configuration object mappers, making the system flexible and extensible instead of hardcoding types

### DIFF
--- a/canvas-server/src/main/java/io/canvasmc/canvas/GlobalConfiguration.java
+++ b/canvas-server/src/main/java/io/canvasmc/canvas/GlobalConfiguration.java
@@ -501,7 +501,7 @@ public class GlobalConfiguration extends Part {
                 "different world due to setup reasoning, like needing to send players to the \"spawn\" world or something.",
                 "This also applies to the end portal and nether portal, in replacement of the overworld, meaning the",
                 "target dimension for entities going from the nether for example will be sent here"
-            ).identifier(); // TODO - object mapping?
+            ).identifier();
     }
 
     public String serverModName = ServerBuildInfo.buildInfo().brandName();
@@ -510,6 +510,10 @@ public class GlobalConfiguration extends Part {
     public boolean cacheMinecraft2BukkitEntityTypeConversion = false;
     public boolean tileEntitySnapshotCreation = false;
     public String defaultRespawnDimensionKey = "minecraft:overworld";
+
+    static {
+        ConfigurationProvider.registerTypeMapper(Identifier.class, Identifier::parse);
+    }
 
     public static @NonNull ResourceKey<@NonNull Level> fetchRespawnDimensionKey() {
         return ResourceKey.create(Registries.DIMENSION, Identifier.parse(GlobalConfiguration.getInstance().defaultRespawnDimensionKey));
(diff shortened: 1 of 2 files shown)


canvas-server/src/main/java/io/canvasmc/canvas/GlobalConfiguration.java
```diff
@@ -501,18 +501,18 @@ public static class Networking extends Part {
...
+            ).identifier();
     }
 
     public String serverModName = ServerBuildInfo.buildInfo().brandName();
     public boolean restoreVanillaEnderPearlBehavior = false;
     public boolean displayWorldLoadScreenForPortaling = true;
     public boolean cacheMinecraft2BukkitEntityTypeConversion = false;
     public boolean tileEntitySnapshotCreation = false;
-    public String defaultRespawnDimensionKey = "minecraft:overworld";
+    public ResourceKey<Level> defaultRespawnDimensionKey = Level.OVERWORLD;
```
Dueris: Is this safe to load here? IIRC we load the config during block registration since that's the earliest things are loaded, but Im curious if this is still fine to call or if by calling this it will load other things and registries that aren't quite ready yet

A good way to test is just a runtime test so if uve runtime tested this and it booted fine(only boot is needed since this would only happen at boot time) then Im alr with this

canvas-server/src/main/java/io/canvasmc/canvas/configuration/ConfigurationProvider.java
```diff
@@ -42,7 +47,16 @@ public class ConfigurationProvider {
...
 
+    private static final Map<Class<?>, Function<String, Object>> TYPE_MAPPERS = new HashMap<>();
+
+    public static <T> void registerTypeMapper(Class<T> clazz, Function<String, T> mapper) {
+        TYPE_MAPPERS.put(clazz, (Function<String, Object>) mapper);
+    }
+
     static {
+        registerTypeMapper(Identifier.class, Identifier::parse);
+        registerTypeMapper(ResourceKey.class, (value) -> ResourceKey.create(Registries.DIMENSION, Identifier.parse(value)));
```
Dueris: Im not a fan entirely of having these 2 registries in here, Id rather have them in the global config file

Also by making resource keys only for the DIMENSION registry, we then make it so we can't really have other types of resource keys. Im curious if maybe we should do some other alternative or if we could find some other way to make that sort of thing work without preventing other resource key typed objects to be used in the future

Dueris (changes requested):

Dueris (approved):

R00tB33rMan (approved):

Toffikk:
@RuleGaed This has been reverted as it causes crashes on startup. Next time please test your changes

RuleGaed:
I actually tested it. What kind of problem did you encounter?

Toffikk:
I got a core dump on startup while it was loading the configuration -> https://mclo.gs/6BUeFkq

Toffikk:
Seems it was related to me having an already existing config file, deleting it fixed the issue but then a stack overflow appeared while regenerating it: https://mclo.gs/SdZ5i0M

RuleGaed:
I understand, there was no problem when I tested it, it might be my mistake.

PR #230, opened 2026-05-15
Restore Scoreboards

Dueris:
This PR restores the scoreboards system from Vanilla Minecraft, disabled by Folia.

The scoreboard system is a technical global state that allows setting integer values to objectives for score holders(entities). This system, while may seem simple to the naked eye, is actually quite complex internally in the terms of trying to make it region-threading safe, so SpottedLeaf disabled this.

This PR reintroduces the scoreboard system through a SWMR format(single write, multiple read), meaning it is safe to write only on 1 thread at a time, but it is thread-safe to read this from multiple threads. The only 2 exceptions to this being the actual housekeeping of player scores, where entry placement/removal are thread-safe alongside reads via a `ConcurrentHashMap`, and the set of score holder names within the `PlayerTeam` class, made into a `CopyOnWriteArraySet`.

> It is worth noting that the "score holder name" for entities is always the string UUID of the entity, or for players, the name associated with their game profile

The "main thread" in this case would be the global tick, which is the **only** thread permitted to writing data to the scoreboard system beyond fetching data and incrementing scores and such. Only on the global tick can you do the following:
- register new teams
- unregister new teams
- update or modify any data associated with teams like the display name, suffix, prefix, color, etc
- add new objectives
- remove objectives
- pack data
- save data
- modify displayed objectives
- update or modify any data associated with objectives like the number format, display name, render type, etc

Things you can do from any thread:
- get the team associated with a score holder
- fetch any data associated with teams like the display name, suffix, prefix, color, etc
- fetch any data associated with objectives like the number format, display name, render type, etc
- fetch the integer value of a score - **Note, the `ScoreAccess` implementation has all of it's methods synchronized now**
- modify the integer value of a score - **Note, the `ScoreAccess` implementation has all of it's methods synchronized now**
- fetch objectives
- get the team associated with the name of a team
- get all team names
- get all team instances
- add score holders to a team
- remove score holders from a team
- probably more that Im forgetting at the time of writing this

There are a few things left to do still, however this is the base implementation that I came up with. The remaining things to do are:
- [x] Restore the `scoreboard` command
- [x] Restore the `team` command
- [x] Restore the `teammsg` command
- [x] Restore the `trigger` command
- [x] Decide what to do about the existing API
- ~~Write a new implementation for `collisions.enablePlayerCollisions` in Paper global, as this setting is *wonky af* and was removed due to this reimplementation of the scoreboards system~~


canvas-server/minecraft-patches/base/0025-Restore-Scoreboards.patch
```diff
@@ -0,0 +1,1007 @@
...
+     private static final Logger LOGGER = LogUtils.getLogger();
+-    private final Object2ObjectMap<String, Objective> objectivesByName = new Object2ObjectOpenHashMap<>(16, 0.5F);
+-    private final Reference2ObjectMap<ObjectiveCriteria, List<Objective>> objectivesByCriteria = new Reference2ObjectOpenHashMap<>();
+-    private final Map<String, PlayerScores> playerScores = new Object2ObjectOpenHashMap<>(16, 0.5F);
+-    private final Map<DisplaySlot, Objective> displayObjectives = new EnumMap<>(DisplaySlot.class);
+-    private final Object2ObjectMap<String, PlayerTeam> teamsByName = new Object2ObjectOpenHashMap<>();
+-    private final Object2ObjectMap<String, PlayerTeam> teamsByPlayer = new Object2ObjectOpenHashMap<>();
++    // Canvas start - region threading
++
++    // TODO - remove collide team rule thing from paper... wtf is this bullshit? -- done
```
Toffikk: what’s wrong with the collide team rule?

Dueris: Honestly, wonky implementation, it uses the teams system and adds all players to a team that is generated by Paper at startup and sets the collision rule based off that

Canvas' existing config is already more customizable than the Paper global implementation, and parts of the config implementation was disabled by Folia already. This patch just *fully* removes it

R00tB33rMan (changes requested):
Conditionally decrement this specific call

canvas-server/src/main/java/io/canvasmc/canvas/util/WeakConcurrentCollection.java
```diff
@@ -0,0 +1,186 @@
...
+
+    @Override
+    public boolean remove(final Object o) {
+        if (o == null) return false;
+        for (WeakReference<E> ref : backed) {
+            E value = ref.get();
+            if (o.equals(value)) {
+                ref.clear();
+                backed.remove(ref);
+                liveCount.decrementAndGet();
```
R00tB33rMan: Conditionally decrement (via backed remove)

canvas-server/src/main/java/io/canvasmc/canvas/util/WeakConcurrentCollection.java
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * @author dueris
+ */
+public class WeakConcurrentCollection<E> implements Collection<E> {
+    private final CopyOnWriteArrayList<WeakReference<E>> backed = new CopyOnWriteArrayList<>();
+    private final AtomicInteger liveCount = new AtomicInteger(0);
+
+    @Override
+    public int size() {
+        return Math.max(0, liveCount.get());
```
R00tB33rMan: It'd likely be best to follow CraftBukkit semantics here, leveraging this against a backed `WeakReference`

canvas-server/minecraft-patches/base/0025-Restore-Scoreboards.patch
```diff
@@ -0,0 +1,1927 @@
...
+         }
+     }
+diff --git a/net/minecraft/server/commands/ScoreboardCommand.java b/net/minecraft/server/commands/ScoreboardCommand.java
+index a4fa358bf6bb3c84a024620ee77f90b6ac68089b..4720c90bd6e84a272538bd117e5f7b2331a59a96 100644
+--- a/net/minecraft/server/commands/ScoreboardCommand.java
++++ b/net/minecraft/server/commands/ScoreboardCommand.java
+@@ -465,9 +465,25 @@ public class ScoreboardCommand {
+         int result = 0;
+ 
+         for (ScoreHolder target : targets) {
```
R00tB33rMan, line 266: Console should be able to invoke `/scoreboard players operation`

R00tB33rMan (changes requested):

devnootnoot:
I'm just going to drop a general review rather than reviewing every class on it's own;

You mentioned that scoreboards allow setting integer values to objectives for score holders “(entities)”, but vanilla score holders are not limited to entities. Vanilla scoreboards support fake/non-entity score holders, and a lot of datapacks, command blocks, plugins, and server setups rely on those for global counters, timers, state flags, etc. 

That part seems important to preserve. If entity-backed score holders need special region scheduling, I think that should be handled separately from fake holders instead of rejecting fake holders entirely. Not sure if this is already implemented, or planned, but thought I should mention it.

Another big concern is command result semantics. A few command handlers seem to schedule the actual scoreboard mutation and then immediately return `0`. That is not really equivalent to vanilla command execution. Brigadier return values matter for command blocks, functions, conditional execution, success counts, and error propagation. The command should not report completion before the actual work has happened, and failures should not happen later out-of-band. If a scoreboard command needs to hop to the global tick or a region thread, it still needs some way to preserve the normal command completion/result behavior.

I am also a bit worried about the scoreboard data snapshot model. You mentioned that team/objective data can be fetched from any thread, but `ObjectiveData` and `TeamData` are mutable classes with plain fields, and their setters mutate the same object in place. If those objects are read across threads, that does not seem like a safe immutable snapshot. A reader could potentially observe a partially updated logical state, like a new display name with an old formatted display name, or a partially changed team state. I think these should either be immutable snapshots that get replaced as a whole, or all reads/writes should be protected by the same synchronization mechanism.

`ReadOnlyScore` also seems questionable to me. It prevents mutation through that wrapper, but it still reads from the underlying mutable `Score`. You mentioned that integer score values can be fetched and modified from any thread, and that the `ScoreAccess` implementation has synchronized methods. If other read paths do not use that same synchronization, then they are not protected by that guarantee. Reads through `ReadOnlyScore` should probably either use the same synchronization as writes or return an immutable snapshot of the score state. “Read-only” is not automatically the same as “thread-safe”.

Saving/packing score data may have a similar issue. You said packing/saving must happen on the global tick, but score values may also be modified from any thread. That means saving could still observe score state while it is being changed elsewhere unless score data is snapshotted or protected by the same lock used for score mutations. I think the save path needs a clear consistency model here.

`WeakConcurrentCollection` also looks like it may have a `liveCount` issue. `remove(Object)` only decrements the count if the backing remove succeeds, which is good, but `removeAll` and `retainAll` decrement after calling `backed.remove(ref)` without checking whether the remove actually succeeded. With concurrent callers, that could make `liveCount` drift and become inaccurate. Those paths should probably only decrement when the backing collection actually removed the reference.

A lower-priority concern is the player-to-team lookup. Removing the direct lookup and replacing it with scanning teams can make `getPlayersTeam(name)` much more expensive, especially since team lookups can be called from hot paths like name formatting, collisions, commands, and plugin APIs. A concurrent player-to-team index might preserve the old behavior better, assuming it can be kept in sync safely with membership changes.

The command error handling also feels a bit inconsistent right now. Some paths throw, some send a failure message, and some schedule work and may send failures later. That can change how commands behave in functions and command blocks. I think the scoreboard/team command implementations should try to preserve vanilla-style exception/result behavior as closely as possible.

Sorry for the long read haha @Dueris

Dueris:
> You mentioned that scoreboards allow setting integer values to objectives for score holders “(entities)”, but vanilla score holders are not limited to entities. Vanilla scoreboards support fake/non-entity score holders, and a lot of datapacks, command blocks, plugins, and server setups rely on those for global counters, timers, state flags, etc.

Score holders when entered into the scoreboard system are the assigned scoreboard name, which for players is the game profile name, and for entities is their UUID. If for example we run the command:

```
/team join test *
```

With the assumption that the team `test` is already created, in Vanilla this won't really do anything, since when searching for teams and adding to teams, etc, it searches directly against the name, assuming it's the profile name or UUID. Meaning if you run this and for some reason it passes, but then try and check later for the team of an entity supposedly in the wildcard, it just returns `null`, since it has to be a lookup of the specific scoreboard name. Entity arguments, via commands, are handled specifically so they all return entity objects when run.

Ive modified locally to make it so it *can* support this via the command, since technically Canvas only really supported if the entity was existing in the server via the command.

For the API, Ive also removed my guards against using the String provided methods for setting/creating score data. That was put in place because technically it could be used to bypass thread checks for players. Even though the system itself is safe to execute concurrently, it isn't really recommended.

Dueris:
> That part seems important to preserve. If entity-backed score holders need special...

Yeah `Entity` score holders are thread checked in a few spots, not all tho.  Anything that isn't an entity isn't checked and can technically be run on any thread. Which does leave room for plugins to intentionally get around thread checks, which is *fine*, the thread checks are more for trying to keep consistency of the entity in question, it won't actually break anything since the system itself is thread-safe for those points we check the entity

> Another big concern is command result semantics. A few command handlers seem to schedule...

**Both** Folia and Canvas sorta ignore this, since the primary functions that actually read this value aren't even enabled in region threading anymore. It's sorta sad, but not something we can really reimplement without blocking of *some* sorts, or a different command system entirely.

> I am also a bit worried about the scoreboard data snapshot model. You mentioned that team/objective...

Yeah this is intentional. The class is updated via an atomic reference if I remember correctly, all set/get operations go through an atomic ref. However I am open to changing this up a bit

> ReadOnlyScore also seems questionable to me. It prevents mutation through that wrapper...

Yes this was more made in a way to sorta just prevent people from getting said read only score and then just casting it to a `Score` instance and just being able to modify it. Sorta trying to enforce the "read only" bit. I changed the `ReadOnlyScore` class to be an immutable snapshot though

> Saving/packing score data may have a similar issue. You said packing/saving must happen on...

While I agree, we also can't entirely just... do locking... since that would lock *everything* from start to finish of the packing/saving, which isn't very ideal in an environment like Folia. This is auto saved on the normal auto save interval, and on `save-all`, and on the end of the server lifecycle. However it is worth noting that ***ONLY*** player scores will potentially have the chance of missing a value or 2 if modified at the exact time of save. Everything else though must be written on the global tick though, so that won't encounter issues. There *are* ways around this like potentially synchronizing or locking the ability to get/write score values, however this then creates a sort of performance burden which isn't very ideal when to solve a very rare race condition that won't actually do much but not save a value entered on a specific nanosecond or 2, when it's covered on shutdown and autosave anyway.

> WeakConcurrentCollection also looks like it may have a liveCount issue. remove(Object) only...

Nice catch, fixed now.

> A lower-priority concern is the player-to-team lookup. Removing the direct lookup and...

Agreed, Ill see if I can do something like this

> The command error handling also feels a bit inconsistent right now. Some paths throw, some...

Command blocks and functions are disabled in Folia and Canvas, and will not be returning probably. As explained above, this is how Canvas and Folia do this to make things region-threading safe.

devnootnoot:
Took a look at the changes you made, and I think most of my previous concerns are now fixed.

The `ReadOnlyScore` immutable snapshot change looks good, the `WeakConcurrentCollection` count issue looks fixed, and I understand the command result semantics point better now. If command blocks/functions are disabled and preserving Brigadier result values would require blocking or a different command system, then I would not treat that as a blocker. Might still be worth documenting somewhere, but I get why it works this way in this environment.

Same with save/pack: I still think there is technically a race where a score value modified at the exact same time as saving may not be captured, but your explanation makes sense. If only player score values can be missed in that tiny window, and everything structural is global-tick-only, then this feels ok to me.

The one concrete thing I would still check is the new team lookup/index behavior. In `addPlayerToTeam`, when a holder is already in another team, I see the holder mapping being updated to the new team, but I do not clearly see the holder being removed from the old team’s actual `players` set before being added to the new one.

If that is the case, the old team’s `getPlayers()` could still contain the holder while `getPlayersTeam(holder)` returns the new team from the lookup map. That would reintroduce a consistency issue, just in a different form.

I think the swap/update block should preserve the vanilla behavior of “remove from old team first, then add to new team”, but do it inside the same safe update path used for the team reference/index. If this is already handled somewhere else and I missed it, then this point can be ignored.

The only broader design thing I am still unsure about is `ObjectiveData`/`TeamData`. I understand that access goes through an atomic reference, but if the same referenced object is mutated in place, the atomic ref does not by itself make that object an immutable snapshot. If updates replace the whole data object, then great. If not, I still think immutable replacement would fit the model better than mutable plain fields.

Dueris:
Nice catch with `addPlayerToTeam`, fixed now

devnootnoot:
The only small thing I would still consider is making `ObjectiveData` and `TeamData` harder to misuse in the future. They are still mutable classes with setters, so the safety now depends on everyone continuing to use this pattern and never mutating the active referenced object directly. That is fine with the current usage, but making them immutable records/classes, or at least reducing setter visibility, would make the model harder to accidentally break later. Just a code quality thing.

Not a blocker from me though. Just a suggestion. Overall looks good to me now!

devnootnoot (approved):

R00tB33rMan (approved):

PR #235, opened 2026-05-20
Log Cleaner

Dueris:
This implements a log cleaner system to the start of the configuration system at startup that cleans old log files if enabled.

This introduces 3 options to the global configuration:
```yaml
logs:
   # Auto-removes log files from the "logs" directory based on how old they are
   enable-log-cleaner: false
   # The amount of the time unit until the log file is marked for deletion
   length: 30
   # The type of time unit to use when comparing how old the file is to the current time
   unit: "DAYS"
```
This allows server owners to sorta clear their logs directory over time, which can be beneficial so it doesn't fill up disk space with logs from ages ago that nobody cares about anymore. The default is disabled with a time threshold of 30 days


Toffikk:
this would be a better fit for a plugin tbh but lgtm

Toffikk (approved):

R00tB33rMan (approved):
Although I partially agree with the above approval comment, it's also nice to have something automatically managed by the server software itself, since the server software also supplies the entire logging feature. Makes it easier to ship logging cleaners without the need for extra maintenance

Issue #234, opened 2026-05-19
Desynchronized day cycle


MrNicolasH:
The day cycle between the client and the server becomes desynchronized, making it night from the client's perspective, but the server's logic shows day ticks.

https://imgur.com/a/hors3PN

Canvas 26.1.2 BUILD 756

Dueris:
Is this replicatable in Folia

I dont think Canvas adjusts the day/night synchronization

Dueris:
Any updates on this?

MrNicolasH:
Yes, it was due to low TPS, caused by poor configuration. My Canvasmc server didn't have all 10 cores of my machine allocated, and when more than 60 players were playing in my SMP, it would become desynchronized.

Dueris:
So this IS or ISNT a Canvas issue?

Toffikk:
If this is indeed a Canvas issue, please comment here and we’ll reopen this

PR #236, opened 2026-05-21
Fix desynchronized day cycle (#234)

RuleGaed:
forceGameTimeSynchronization()` was sending an empty `Map.of()` for clock updates. Replaced with `this.clockManager().createFullSyncPacket()` so the client receives proper `WorldClock` states


Toffikk:
Please DO NOT hand edit patches like this. They won’t apply

PR #243, opened 2026-05-28
fix: restore CANVAS brand compatibility in isBrandCompatible

Euphillya:
This fix resolves this following bug: 

```java
    @Override
    public boolean isAvailable() {
        var BRAND_CANVAS_ID = Key.key("canvasmc", "canvas");
        var buildInfo = ServerBuildInfo.buildInfo();
        boolean isCanvas = buildInfo.isBrandCompatible(BRAND_CANVAS_ID);
        log.info("Use CANVAS : {}", isCanvas);
        return isCanvas;
    }
```

Response without the fix : 
[15:56:41 INFO]: [fr.euphyllia.skyllia.hook.canvas.CanvasHook] Use CANVAS : false


Dueris (approved):

Toffikk:
This is wrong, as `this.brandId` is set to the value of `Brand-Id` in the manifest which matches the key of `canvasmc:canvas` and running your example locally, without this fix, worked for me
image

Toffikk:
image
for reference

PR #245, opened 2026-05-28
fix: ensure waypoints respect API

### LINKED ISSUE #244: Locator bar does not respect certain API methods
The locator bar on Canvas does not hide players from it when calling `Player#hidePlayer(Plugin, Player)` which differs from the normal behavior on Paper servers.

Tested on Canvas 26.1.2 build 775

Jsinco:
closes #244

### DIFF
--- a/canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
+++ b/canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
@@ -6011,15 +6011,16 @@ index 67415351783d15aa97b011967aa9067c899e6031..18004deafdb0ce15a9fc50ae49d34c40
          .comapFlatMap(
              name -> byName(name).map(DataResult::success).orElse(DataResult.error(() -> "No scoreboard criteria with name: " + name)),
 diff --git a/net/minecraft/world/waypoints/WaypointTransmitter.java b/net/minecraft/world/waypoints/WaypointTransmitter.java
-index 2866b750a28cc32c2913d3c1afee95f05956982c..a21eed9e39f665588500acf2e89142a14d8ae3aa 100644
+index 2866b750a28cc32c2913d3c1afee95f05956982c..7f9cbda2d3e86acf8d145ea6bddbf99869d2d34a 100644
 --- a/net/minecraft/world/waypoints/WaypointTransmitter.java
 +++ b/net/minecraft/world/waypoints/WaypointTransmitter.java
-@@ -20,20 +20,35 @@ public interface WaypointTransmitter extends Waypoint {
+@@ -20,20 +20,36 @@ public interface WaypointTransmitter extends Waypoint {
      Waypoint.Icon waypointIcon();
  
      static boolean doesSourceIgnoreReceiver(final LivingEntity source, final ServerPlayer receiver) {
 -        if (!receiver.getBukkitEntity().canSee(source.getBukkitEntity())) return true; // Paper - ignore if entity is hidden from player
 +        // Canvas start - region threading
++        if (!receiver.getBukkitEntity().canSeeChunkMapUpdatePlayer(source.getBukkitEntity())) return true; // Paper - ignore if entity is hidden from player
          if (receiver.isSpectator()) {
              return false;
 -        } else if (!source.isSpectator() && !source.hasIndirectPassenger(receiver)) {


Dueris (changes requested):
This needs to be merged into `Fixup-Region-Threading`

Jsinco:
cleaned up

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
```diff
@@ -6011,15 +6011,16 @@ index 67415351783d15aa97b011967aa9067c899e6031..18004deafdb0ce15a9fc50ae49d34c40
...
-index 2866b750a28cc32c2913d3c1afee95f05956982c..a21eed9e39f665588500acf2e89142a14d8ae3aa 100644
+index 2866b750a28cc32c2913d3c1afee95f05956982c..d30cb3b96c6bee57063c256c28cf149034cd7be0 100644
 --- a/net/minecraft/world/waypoints/WaypointTransmitter.java
 +++ b/net/minecraft/world/waypoints/WaypointTransmitter.java
-@@ -20,20 +20,35 @@ public interface WaypointTransmitter extends Waypoint {
+@@ -20,20 +20,36 @@ public interface WaypointTransmitter extends Waypoint {
      Waypoint.Icon waypointIcon();
  
      static boolean doesSourceIgnoreReceiver(final LivingEntity source, final ServerPlayer receiver) {
 -        if (!receiver.getBukkitEntity().canSee(source.getBukkitEntity())) return true; // Paper - ignore if entity is hidden from player
```
Toffikk, line 6021: can't this removal just be reverted here? no need to move that line down with a different comment here

Toffikk (changes requested):

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
Jsinco: done

Dueris:
`canSee` is a generally expensive method to be conducting in practice. I am unsure how often this method is called off the top of my head, but I think it'd be worth considering trying to find a place in this method where we can try and call this as little as possible while maintaining compatibility with Paper's waypoint system.

Dueris:
```java
    static boolean doesSourceIgnoreReceiver(final LivingEntity source, final ServerPlayer receiver) {
        // Canvas start - region threading
        if (receiver.isSpectator()) {
            // here we should
            return false;
        }

        if (!source.isSpectator()) {
            double dx = source.getX() - receiver.getX();
            double dz = source.getZ() - receiver.getZ();
            double horizontalDistSq = dx * dx + dz * dz;

            if (horizontalDistSq < 4.0) {
                if (source.hasIndirectPassenger(receiver)) {
                    // here we don't need to since we don't track indirect passengers anyway
                    return true;
                }
            }

            double transmitRange = source.getAttributeValue(Attributes.WAYPOINT_TRANSMIT_RANGE);
            double receiveRange = receiver.getAttributeValue(Attributes.WAYPOINT_RECEIVE_RANGE);
            double minRange = Math.min(transmitRange, receiveRange);

            // here we should
            return source.distanceTo(receiver) >= minRange;
        }

        return true;
        // Canvas end - region threading
    }
```

I marked spots where we could add this check. Idk, it probably is fine to keep it at the head of the method, but that's my 2 cents. Im fine with this PR being merged though, just wanting to throw in if there is a better place we should put this

Toffikk:
Tbh i think the current diff is good too, as even if that method is a bit expensive that’s what paper does anyway and we optimize waypoints to the point of this not being that relevant

Dueris:
I don't think the same entity can be passed as the receiver and source, yes? I ask bc I'm not on my computer rn

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
```diff
@@ -6011,14 +6011,13 @@ index 67415351783d15aa97b011967aa9067c899e6031..18004deafdb0ce15a9fc50ae49d34c40
...
+index 2866b750a28cc32c2913d3c1afee95f05956982c..f8b6d8bf6851a025f95cfe8328b785266218f2ed 100644
 --- a/net/minecraft/world/waypoints/WaypointTransmitter.java
 +++ b/net/minecraft/world/waypoints/WaypointTransmitter.java
-@@ -20,20 +20,35 @@ public interface WaypointTransmitter extends Waypoint {
-     Waypoint.Icon waypointIcon();
+@@ -21,19 +21,35 @@ public interface WaypointTransmitter extends Waypoint {
  
      static boolean doesSourceIgnoreReceiver(final LivingEntity source, final ServerPlayer receiver) {
--        if (!receiver.getBukkitEntity().canSee(source.getBukkitEntity())) return true; // Paper - ignore if entity is hidden from player
+         if (!receiver.getBukkitEntity().canSee(source.getBukkitEntity())) return true; // Paper - ignore if entity is hidden from player
```
Dueris: We can use `canSeeChunkMapUpdatePlayer`, instead of `canSee`. The source and receiver should never be the same, and this check will remove the `equals` comparison, which is better than `canSee`

Dueris (changes requested):

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
Jsinco: done ^^

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
```diff
@@ -6011,14 +6011,15 @@ index 67415351783d15aa97b011967aa9067c899e6031..18004deafdb0ce15a9fc50ae49d34c40
...
+index 2866b750a28cc32c2913d3c1afee95f05956982c..484ade618d48017b6098bcf55bc007a00a6381a7 100644
 --- a/net/minecraft/world/waypoints/WaypointTransmitter.java
 +++ b/net/minecraft/world/waypoints/WaypointTransmitter.java
-@@ -20,20 +20,35 @@ public interface WaypointTransmitter extends Waypoint {
+@@ -20,20 +20,36 @@ public interface WaypointTransmitter extends Waypoint {
      Waypoint.Icon waypointIcon();
  
      static boolean doesSourceIgnoreReceiver(final LivingEntity source, final ServerPlayer receiver) {
 -        if (!receiver.getBukkitEntity().canSee(source.getBukkitEntity())) return true; // Paper - ignore if entity is hidden from player
++        if (!receiver.getBukkitEntity().canSeeChunkMapUpdatePlayer(source.getBukkitEntity())) return true; // Canvas - respect vanish API
```
Dueris: Could we move this down into the `// Canvas start - region threading` bit, and restore the Paper comment

Dueris (changes requested):

Dueris:
That's my only final nitpick, once that's done Im fine with merging

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
Dueris: And remove Canvas' `Canvas - respect vanish API` comment too

MrMasrozYTLIVE (approved):
LGTM

Jsinco:
accidentally closed, should be fine to merge now after the edits

Dueris (approved):

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
Toffikk: > We can use `canSeeChunkMapUpdatePlayer`, instead of `canSee`. The source and receiver should never be the same, and this check will remove the `equals` comparison, which is better than `canSee`

Eh, while this would be true in a perfect world, some plugins do that and i don’t think we can rely on ideal practice. The equals check was added in that method specifically because of such uses. https://hub.spigotmc.org/jira/browse/SPIGOT-7312
I’m personally for keeping the original `canSee` call

Toffikk: but this is also in NMS so maybe it’s fine to bypass it

Toffikk (approved):

Toffikk:
If any issues arise we can always fix them up

PR #246, opened 2026-05-28
Revert "fix: restore CANVAS brand compatibility in isBrandCompatible"

Toffikk:
Reverts CraftCanvasMC/Canvas#243

TODO: wait for euphyllia's response before potentially reverting that commit


Euphillya:
Strange because my Skyblock plugin returns false. But it seems logical, I don't understand

Euphillya:
Oh no, I think I know why. I'll check tomorrow, but if it's what I think it is, then it's definitely not a bug.

Euphillya:
@Toffikk you can revert, I literally confused it with this: https://github.com/CraftCanvasMC/Baguette/blob/e3547a54a68401a988225a2cb9e5b12e7f612f75/baguette-server/paper-patches/files/src/main/java/io/papermc/paper/ServerBuildInfoImpl.java.patch#L28

Euphillya (approved):

PR #228, opened 2026-05-13
World key migration of Canvas’s internals

Toffikk:

PR #248, opened 2026-05-29
Matching Folia Regions on Spark Report

Euphillya:
This PR updates the World tab of the Spark website to display Folia regions with their location center instead of those generated by Spark, which can be confusing.

### DIFF
--- a/canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
+++ b/canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
@@ -141,37 +141,34 @@ public class FoliaWorldInfoProvider implements WorldInfoProvider {
             else return new ChunksResult<>(); // global tick, doesn't own chunks
         }
 
-        final long now = System.nanoTime();
         for (World world : this.server.getWorlds()) {
             final ServerLevel level = ((CraftWorld) world).getHandle();
             final Map<Long, List<FoliaChunkInfo>> byRegion = new HashMap<>();
-            final Map<Long, RegionStats> statsByRegion = new HashMap<>();
+            final Map<Long, ChunkRegionCenter> centerByRegion = new HashMap<>();
 
             for (Chunk chunk : world.getLoadedChunks()) {
                 if (chunk == null) continue;
-                final RegionStats stats = FoliaChunkInfo.resolveRegionStats(level, chunk.getX(), chunk.getZ(), now, statsByRegion);
-                final FoliaChunkInfo info = new FoliaChunkInfo(chunk, world, this.plugin, stats);
-                byRegion.computeIfAbsent(stats.id(), k -> new ArrayList<>()).add(info);
+                final ChunkRegionCenter center = FoliaChunkInfo.resolveRegionCenter(level, chunk.getX(), chunk.getZ(), centerByRegion);
+                final FoliaChunkInfo info = new FoliaChunkInfo(chunk, world, this.plugin, center);
+                byRegion.computeIfAbsent(center.id(), k -> new ArrayList<>()).add(info);
             }
 
             for (Map.Entry<Long, List<FoliaChunkInfo>> entry : byRegion.entrySet()) {
-                final RegionStats stats = statsByRegion.getOrDefault(entry.getKey(), RegionStats.UNKNOWN);
-                data.put(buildRegionKey(world.getName(), stats), entry.getValue());
+                final ChunkRegionCenter center = centerByRegion.getOrDefault(entry.getKey(), ChunkRegionCenter.UNKNOWN);
+                data.put(buildRegionKey(world.getName(), center), entry.getValue());
             }
         }
 
         return data;
     }
 
-    private static String buildRegionKey(String worldName, RegionStats stats) {
-        if (stats.id() == -1L) {
+    private static String buildRegionKey(String worldName, ChunkRegionCenter center) {
+        if (center.id() == -1L) {
             return worldName + " [region=unknown]";
         }
 
-        return worldName
-            + " [ TPS: " + String.format("%.1f", stats.tps())
-            + " - MSPT: " + String.format("%.2f", stats.mspt()) + "ms"
-            + " - Util: " + String.format("%.2f", stats.util() * 100.0) + "% ]";
+        final String worldCapitalized = worldName.substring(0, 1).toUpperCase(Locale.ROOT) + worldName.substring(1);
+        return worldCapitalized + " region at " + center.centerBlockX() + ", " + center.centerBlockZ();
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


Dueris:
My only thought about this is is it worth it to differentiate ourselves in this manner? Like yes, this is a cool concept and I like the idea, but I feel like breaking Sparks formatting of it's provided data may not be the *greatest* idea

Euphillya:
In itself, I'm not changing how Spark works. When Spark sends chunk groups, it's to avoid displaying a huge list of entities, chunks, etc. By changing how Spark generates its sections using Folia regions, it will allow both developers and server administrators to have a better representation.
I don't know if I've managed to make my point clear.

Dueris:
Ah gotcha, that makes more sense. Im fond of the idea, Ill discuss in our GC(the team) later regarding this and then Ill conduct my review code-wise

canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
```diff
@@ -136,45 +141,55 @@ public ChunksResult<FoliaChunkInfo> pollChunks() {
...
+            for (Map.Entry<Long, List<FoliaChunkInfo>> entry : byRegion.entrySet()) {
+                final RegionStats stats = statsByRegion.getOrDefault(entry.getKey(), RegionStats.UNKNOWN);
+                data.put(buildRegionKey(world.getName(), stats), entry.getValue());
+            }
         }
 
         return data;
     }
 
+    private static String buildRegionKey(String worldName, RegionStats stats) {
```
Dueris: After discussing with the team, we thought it'd be better if the name would be something like:

`{world, first letter capitalized} region at {center block of center chunk of the region, leave out the "y" value}`

Dueris (changes requested):
To go along with this, Im not quite sure if showing the TPS, MSPT, and utilization is really necessary

canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
Euphillya: image
image

canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
```diff
@@ -192,16 +210,39 @@ public Collection<DataPackInfo> pollDataPacks() {
...
+        private static ChunkRegionCenter resolveRegionCenter(@NonNull ServerLevel level, int chunkX, int chunkZ, @NonNull Map<Long, ChunkRegionCenter> cache) {
+            final var region = level.regioniser.getRegionAtSynchronised(chunkX, chunkZ);
+
+            if (region == null) return ChunkRegionCenter.UNKNOWN;
+            final long id = region.id;
+            final ChunkRegionCenter cached = cache.get(id);
+            if (cached != null) return cached;
+
+            final ChunkPos center = region.getCenterChunk();
+            final int centerBlockX = center != null ? center.x() * 16 + 8 : 0;
```
Dueris: Use `getMiddleBlockX`

canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
```diff
@@ -192,16 +210,39 @@ public Collection<DataPackInfo> pollDataPacks() {
...
+            final int centerBlockZ = center != null ? center.z() * 16 + 8 : 0;
```
Dueris: Use `getMiddleBlockZ`

canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
```diff
@@ -192,16 +210,39 @@ public Collection<DataPackInfo> pollDataPacks() {
...
+
+            final ChunkRegionCenter regionCenter = new ChunkRegionCenter(id, centerBlockX, centerBlockZ);
```
Dueris: I think if the center chunk is null then we should probably just not include this region, since if the center chunk is null it has no loaded chunks, which then we gotta ask, if there are no loaded chunks, then why would this have any data

And also, falling back to `0,0` for this is a bad idea, because we could then have multiple regions that are "dead" displaying `0,0` as their region center, which makes no sense

canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
```diff
@@ -192,16 +210,39 @@ public Collection<DataPackInfo> pollDataPacks() {
...
+
+        FoliaChunkInfo(@NonNull Chunk chunk, World world, FoliaSparkPlugin plugin, ChunkRegionCenter center) {
+            super(chunk.getX(), chunk.getZ(), center);
 
             Executor executor = task -> RegionizedServer.getInstance().taskQueue.queueTickTaskQueue(((CraftWorld) world).getHandle(), getX(), getZ(), task, Priority.BLOCKING);
             this.entityCounts = CompletableFuture.supplyAsync(() -> calculate(chunk), executor);
         }
 
+        private static ChunkRegionCenter resolveRegionCenter(@NonNull ServerLevel level, int chunkX, int chunkZ, @NonNull Map<Long, ChunkRegionCenter> cache) {
+            final var region = level.regioniser.getRegionAtSynchronised(chunkX, chunkZ);
```
Dueris: Unsure if we should use synchronized or not here, since this executes on every loaded chunk

canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
```diff
@@ -136,23 +141,36 @@ public ChunksResult<FoliaChunkInfo> pollChunks() {
...
+                if (chunk == null) continue;
+                final ChunkRegionCenter center = FoliaChunkInfo.resolveRegionCenter(level, chunk.getX(), chunk.getZ(), centerByRegion);
+                final FoliaChunkInfo info = new FoliaChunkInfo(chunk, world, this.plugin, center);
+                byRegion.computeIfAbsent(center.id(), k -> new ArrayList<>()).add(info);
             }
 
-            data.put(world.getName(), list);
+            for (Map.Entry<Long, List<FoliaChunkInfo>> entry : byRegion.entrySet()) {
+                final ChunkRegionCenter center = centerByRegion.getOrDefault(entry.getKey(), ChunkRegionCenter.UNKNOWN);
+                data.put(buildRegionKey(world.getName(), center), entry.getValue());
```
Dueris: Should we use the Bukkit name, or the Minecraft dimension identifier path?

Bukkit - Minecraft
`world` - `overworld`
`world_nether` - `the_nether`
`world_the_end` - `the_end`

canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
```diff
@@ -192,16 +210,39 @@ public Collection<DataPackInfo> pollDataPacks() {
...
     }
 
-    public static final class FoliaChunkInfo extends AbstractChunkInfo<EntityType> {
+    public static final class FoliaChunkInfo extends AbstractFoliaChunkInfo<EntityType> {
         private final CompletableFuture<CountMap<EntityType>> entityCounts;
 
         FoliaChunkInfo(@NonNull Chunk chunk, World world, FoliaSparkPlugin plugin) {
-            super(chunk.getX(), chunk.getZ());
+            this(chunk, world, plugin,
+                resolveRegionCenter(((CraftWorld) world).getHandle(), chunk.getX(), chunk.getZ(), new HashMap<>()));
```
Dueris: Collapse this new line or split it so its:
```
this(
    chunk,
    world,
    plugin,
    etc...
);
```

canvas-server/src/main/java/me/lucko/spark/paper/common/platform/world/WorldStatisticsProvider.java
```diff
@@ -0,0 +1,216 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class WorldStatisticsProvider {
```
Dueris, line 20: Can you add `// Canvas` documentation like a normal patch for where things were modified from the original?

And for added imports, sorta the same deal as a normal patch, FQCN

Dueris (changes requested):

canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
Euphillya: It depends on what we want: either keep the format as Spark does (even though we now use a capital letter) or use Minecraft dimension identifier path.

Dueris: I personally like the Minecraft name, makes more sense IMO, I dont rly like the Bukkit name ever 😂

Euphillya: That's changed: 

Overworld : https://spark.lucko.me/yrEmgHH4p5
The_nether: https://spark.lucko.me/tYtMH6MATQ

Dueris (approved):

Dueris:
Perhaps update the PR description for the other people that are reviewing this later on

canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
```diff
@@ -145,6 +145,8 @@ public ChunksResult<FoliaChunkInfo> pollChunks() {
             final ServerLevel level = ((CraftWorld) world).getHandle();
             final Map<Long, List<FoliaChunkInfo>> byRegion = new HashMap<>();
             final Map<Long, ChunkRegionCenter> centerByRegion = new HashMap<>();
+            final String worldKey = world.getKey().getKey();
```
Toffikk: use `world.key().asString()`

Toffikk (changes requested):

canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
Euphillya: asString returns: namespace:world
You should check with @Dueris which one you prefer.

Toffikk: I think the key().asString() is more preferred since that what all of our internals use rn, #228

Toffikk: but wait for him to reply

Dueris: Yah but like, would we rather an identifier for Spark or the dimension key.

The dimension key makes more sense, removes the redundant `minecraft:` at the beginning

Toffikk: then perhaps world.key().asMinimalString() ?
It removes the namespace if it is the minecraft one

Dueris: Oh really? Yeah Id be down for that. But with that, replace it with the same pattern as the Identifier debug file name:

`str.replace('/', '_').replace(':', '_')`

Euphillya: I changed

Dueris (approved):

canvas-server/src/main/java/io/canvasmc/canvas/spark/plugin/FoliaWorldInfoProvider.java
Toffikk: could you also change getKey() to key() ? so that we operate on the adventure one

Euphillya: good

Toffikk (approved):

PR #249, opened 2026-05-31
Improve World Unload API

Dueris:
Canvas' world unload system has been utilizing the region scheduler threads to conduct shutdown and then finalize via the global tick. While this was good in practice, this wasn't entirely the best approach due to the fact that the full region unload would occur on the scheduler thread, which if we have a ton of regions loaded during world unload, we can sorta clog the scheduler threads with unloading regions and then removing the world in the global tick.

Overall, it was a good system, but this is better, since we now move all of this to a single thread extending the `RegionShutdownThread`, meaning we bypass thread checks on this thread(similar to shutdown), of which we conduct a process nearly identical to shutdown of the server, but specifically for the world.

This also improves race condition handling of when players are portaling or teleporting across dimensions into an unloading dimension. While this race condition window was small due to previous implementations, this window is now fully closed off and the race condition should be fixed, given we now track the destination `ServerLocation` during teleporting and portaling, allowing us to determine any players in-flight to the unloading world and send them back to the origin. If a player does not have an origin at all, we boot the player, since login is already handled for when the world doesn't exist or is actively unloading

In this PR, we also disable the teleport events for when we send these players back, since simply, we are trying to roll back the player to the previous location as if the previous teleport/portal was cancelled, so we don't want plugins interfering with this or cancelling this, because then we end up in a very situation where the player is stuck in limbo because of a dumb plugin.

Less important, but this also fixes up the patch diff with some things missing from the Paper-Server version of `Fixup-Region-Threading` patch


R00tB33rMan (approved):
Unless I'm blind, can't find anything in specific that's wrong! LGTM

devnootnoot:
The main thing I would double-check is the delayed rollback path. In `WorldShutdownThread`, if a player is still processing the initial teleport/portal and is not removed yet, the rollback gets queued with `RegionizedServer.getInstance().addTask(...)`. I would make sure that task is guaranteed to run before the world is removed from the server/world maps and before the unload ticket completes. If the unload thread keeps going immediately, there may still be a small window where the world unload finishes before that queued rollback actually runs.

Related to that, `ServerLocation` allows `pos`, `yaw`, and `pitch` to be null for portaling, but `teleportPlayer(...)` does a `checkNotNull(target.pos())`. If rollback targets are always real origin positions, then that is fine, but I think it would be worth making that invariant explicit. Otherwise a portal-style `ServerLocation` reaching the rollback path would throw during world unload, which now stops the server.

devnootnoot:
LGTM

devnootnoot (approved):

R00tB33rMan (approved):
Tested and no more edge-cases!

PR #253, opened 2026-06-05
Fix /tps callback retaining region data

BaconCat1:
Prebuild the expanded `/tps` region details before registering the click callback so the callback does not retain live region data/list objects.

This avoids the callback manager keeping region/tick objects reachable longer than intended.


Dueris:
Merge this into Fixup Region Threading

ZepsiZola:
Amen brother

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
```diff
@@ -1153,6 +1161,13 @@ index 7826c50b7c7c59e148905337ddc0dc8085f3f7c1..37519ae684629e4bc7758fdefb1990d3
  
      @Override
      public boolean execute(final CommandSender sender, final String commandLabel, final String[] args) {
+@@ -360,4 +507,4 @@ public final class CommandServerHealth extends Command {
+         }
```
Dueris: Remove this section added

It does nothing it's just annoying

Dueris (changes requested):

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
BaconCat1: Fixed.

Dueris (approved):

PR #254, opened 2026-06-07
Add unsafe End portal teleportation config for sand duping

LuckyODO:
I feel like many people would need this feature. 

See Luminol


Dueris:
It's called "unsafe" for a reason

I believe SpottedLeaf also commented once about this saying it's unsafe with region threading

Dueris:
This is not something Im comfortable with merging at the time

PR #256, opened 2026-06-09
Add per category entity tick intervals

Nrleryxx:
Each entity category monster creature ambient waterCreature misc can have a separate tick interval 1 means every tick default backward compatible Higher values reduce CPU in mob farms and dense areas Players always tick every tick


coredex-source:
We'd rather not have such a feature in canvas. Thanks for the effort though!

PR #260, opened 2026-06-14
Restore more Vanilla commands

Dueris:
Restores the `data` and `clone` commands

### DIFF
--- a/canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
+++ b/canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
@@ -1256,18 +1256,24 @@ index 8d4f092ec65d285508be81cbfadd6e747e1d067e..2db474ced22d5867b0f6315185af31d3
  
      public static ServerPlayer getPlayer(final String name) {
 diff --git a/net/minecraft/commands/Commands.java b/net/minecraft/commands/Commands.java
-index aa3e2b7aca2163c61de949519f332cd1dba5920a..434563cf7958aa4fec624293968d1236cb351c2e 100644
+index aa3e2b7aca2163c61de949519f332cd1dba5920a..a4d799f297f9178a83b639b2a9d4b9d98d84b2ad 100644
 --- a/net/minecraft/commands/Commands.java
 +++ b/net/minecraft/commands/Commands.java
-@@ -195,7 +195,7 @@ public class Commands {
+@@ -195,11 +195,11 @@ public class Commands {
          AdvancementCommands.register(this.dispatcher);
          AttributeCommand.register(this.dispatcher, context);
          ExecuteCommand.register(this.dispatcher, context);
 -        //BossBarCommands.register(this.dispatcher, context); // Folia - region threading - TODO
 +        BossBarCommands.register(this.dispatcher, context);
          ClearInventoryCommands.register(this.dispatcher, context);
-         //CloneCommands.register(this.dispatcher, context); // Folia - region threading - TODO
+-        //CloneCommands.register(this.dispatcher, context); // Folia - region threading - TODO
++        CloneCommands.register(this.dispatcher, context);
          DamageCommand.register(this.dispatcher, context);
+-        //DataCommands.register(this.dispatcher); // Folia - region threading - TODO
++        DataCommands.register(this.dispatcher);
+         //DataPackCommand.register(this.dispatcher, context); // Folia - region threading - TODO
+         //DebugCommand.register(this.dispatcher); // Folia - region threading - TODO
+         DefaultGameModeCommands.register(this.dispatcher);
 @@ -222,7 +222,7 @@ public class Commands {
          KillCommand.register(this.dispatcher);
          ListPlayersCommand.register(this.dispatcher);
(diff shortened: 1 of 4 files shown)


canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
```diff
@@ -3097,6 +3347,282 @@ index 088b7b91154546fc9c020cab5cefbb4b89762ac8..9d5f76ff2892fd38a21397698a986027
...
+     }
+ 
+     private static int getData(final CommandSourceStack source, final DataAccessor accessor) throws CommandSyntaxException {
+-        CompoundTag data = accessor.getData();
+-        source.sendSuccess(() -> accessor.getPrintSuccess(data), false);
++        // Canvas start - region threading
++        handleAction((dataAccessor) -> {
++            CompoundTag data = dataAccessor.getData();
++            source.sendSuccess(() -> accessor.getPrintSuccess(data), false);
++            return 1;
```
Toffikk, line 3562: note: this should use Command.SINGLE_SUCCESS in 26.2 or now if its available (should be)

Dueris: While I agree, this should be in 26.2 update, since currently the import isnt even in the class, so I dont see a point since that diverges from the rest of our patches formatting and the src for this update

In 26.2 we should handle this

RuleGaed:
CloneCommands java calls setBlock again after loading block entity NBT data this discards the data you just loaded remove the extra setBlock call BlockDataAccessor java isTickThreadFor does not check for null getLevel this throws NPE when entity is unloaded add a null guard like scheduleOrExecute does all data subcommands return 0 now this breaks execute store run data get you already know about this for 26.2 CloneCommands async callbacks have no try catch silent exceptions can leave the world inconsistent scheduleOrExecute silently fails when level is null data storage is still disabled

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
```diff
@@ -1744,6 +1750,250 @@ index fb55ed68a1f8470b551030a27ca2656555c93d9e..5756f7aca1a80782a88b0e82ef82a626
...
++            final ProblemReporter reporter,
++            final net.minecraft.core.HolderLookup.Provider registryAccess
++        ) {
++            if (!predicate.test(block)) {
++                return;
++            }
++
++            boolean last;
++
++            final BlockEntity blockEntity = level.getBlockEntity(pos);
```
Toffikk, line 1793: Could align these to unobf names, blockState -> state, blockEntity i think will be entity or block

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
```diff
@@ -1744,6 +1750,250 @@ index fb55ed68a1f8470b551030a27ca2656555c93d9e..5756f7aca1a80782a88b0e82ef82a626
...
++                        // note: clearBlockList is basically just a list of all the blocks we collected that
++                        //       at least passed the predicate
++
++                        int defaultUpdateFlags = 2 | (strict ? 816 : 0);
++                        if (mode == CloneCommands.Mode.MOVE) {
++                            for (BlockPos pos : snapshot.allPositions) {
++                                fromDimension.setBlock(pos, Blocks.BARRIER.defaultBlockState(), defaultUpdateFlags | Block.UPDATE_SKIP_ALL_SIDEEFFECTS);
++                            }
++
++                            int standardUpdateFlags = strict ? defaultUpdateFlags : Block.UPDATE_ALL;
```
Toffikk, line 1893: `updateFlags`

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
```diff
@@ -3097,6 +3347,282 @@ index 088b7b91154546fc9c020cab5cefbb4b89762ac8..9d5f76ff2892fd38a21397698a986027
...
+     Component getPrintSuccess(NbtPathArgument.NbtPath path, double scale, int value);
++    // Canvas start - region threading
++    com.mojang.brigadier.exceptions.SimpleCommandExceptionType CSE_TARGET_NOT_SAME_REGION =
++        new com.mojang.brigadier.exceptions.SimpleCommandExceptionType(Component.literal("Target not in same region"));
++    // note: this should throw a CSE that says "not in same region" or something,
++    //       we do not trigger a REAL ISE thread check
++    default void ensureTickThread(final String reason) throws CommandSyntaxException {
++        // no-op
++    }
++    default void scheduleOrExecute(final java.util.function.Consumer<DataAccessor> daConsumer) {
```
Toffikk: Should just be consumer

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
```diff
@@ -3097,6 +3347,282 @@ index 088b7b91154546fc9c020cab5cefbb4b89762ac8..9d5f76ff2892fd38a21397698a986027
...
++
++    @Override
++    public void ensureTickThread(final String reason) throws com.mojang.brigadier.exceptions.CommandSyntaxException {
++        if (!isTickThreadFor()) {
++            throw DataAccessor.CSE_TARGET_NOT_SAME_REGION.create();
++        }
++    }
++
++    @Override
++    public void scheduleOrExecute(final java.util.function.Consumer<DataAccessor> daConsumer) {
```
Toffikk: consumer

Toffikk (dismissed):

Toffikk:
Other than that LGTM (tho the comments create a lot of unnecessary diff)

Dueris:
@RuleGaed

> CloneCommands java calls setBlock again after loading block entity NBT data this discards the data you just loaded remove the extra setBlock call

This is to match Vanilla, not a bug, this is also incorrect.

> BlockDataAccessor java isTickThreadFor does not check for null getLevel this throws NPE when entity is unloaded add a null guard like scheduleOrExecute does

It's never null, `scheduleOrExecute` has no such check

> all data subcommands return 0 now this breaks execute store run data get you already know about this for 26.2

This is intentional

> CloneCommands async callbacks have no try catch silent exceptions can leave the world inconsistent

What are you even talking about

> scheduleOrExecute silently fails when level is null data storage is still disabled

This method has no such check, the global data storage is intentionally disabled right now

@Toffikk 
> tho the comments create a lot of unnecessary diff

While I agree, I think putting comments is probably good for general code quality. Better to have a lot of comments explaining what's going on rather than just rawdog code

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
Dueris: `blockEntity` matches Vanilla below

Dueris: Same with `blockState`

canvas-server/minecraft-patches/base/0008-Fixup-Region-Threading.patch
Dueris: Matches Vanilla below

Toffikk (approved):

devnootnoot:
The main thing I would want clearly documented is that this is not fully vanilla-equivalent yet. `StorageDataAccessor` is still disabled, so `/data storage ...` does not work. Also, several /data subcommands now return 0 because the real action may be scheduled onto the owning region. I understand that this is probably intentional and that command blocks/functions/return-value consumers are already limited there, but it should be explicit because vanilla users may expect things like `execute store result run data get ...` to work.


One small concern is the silent no-op style in the accessor scheduling path. For example, if a block/entity accessor cannot schedule because its level is unavailable, the command can effectively do nothing instead of failing loudly. If that state is truly impossible, then it is fine, but I would rather see that invariant enforced or logged than have the command silently disappear.


For clone, I would also just make sure any async/scheduled stage failure cannot leave the source half-moved in MOVE mode. Since source clearing happens before the destination write, a failure during the destination stage could be pretty destructive. If the async callbacks already fail loudly enough and this is accepted behavior, then feel free to ignore.

Dueris:
As for the level being unavailable, that is irrelevant for entities since they are constructed with a `Level` instance and it remains non-null throughout the lifecycle. And scheduling to the entity requires no such level being set. With blocks, it has to be a block *entity*, meaning the block entity needs to be loaded, and it's level is set when loaded in the world.

For the clone cmd, if it fails it crashes, same as Vanilla

As for the `storage` accessor type, that's gonna be documented in Canvas' docs when this is merged. Your cmd for the `execute` cmd also doesnt exist, so this isn't an issue

devnootnoot (approved):

R00tB33rMan (approved):

PR #262, opened 2026-06-16
26.2 tracking

Toffikk:
## Current TODO:
 - [x] update MC source patches - patch roulette
 - [x] update 0008-Fixup-Region-Threading.patch, cc @Dueris 
 - [x] pull #260 in once merged to master - needs to be done before publishing any builds


Dueris:
Once #261 is merged then we can update this branch and Id be fine with builds being out for 26.2

Dueris (approved):

PR #263, opened 2026-06-19
re-implement ChunkAccess#blockEntities for folia internals compat

Jsinco:
I'm creating this PR as a draft because this is a more a matter of opinion rather than a contribution that should, or rather, *needs* to be accepted at all. This draft PR specifically focuses on `ChunkAccess#blockEntities`, I don't believe other fields are as much of a concern.

It is of my opinion that Canvas extends the Folia server. Fixing compatability issues and extending the Folia server back to what is expected on a standard Paper server. Whenever possible, I believe Canvas should introduce optimizations, but not at the cost of breaking compatibility with Folia.

Given how experimental Folia itself is, plugin developers are often required to access internals to achieve the same functionality that is readily available on Paper. Because of this, I believe the removal of `ChunkAccess#blockEntities` is detrimental enough to external plugin development to warrant reintroducing the field for backward compatibility. This patch restores the field while preserving the optimizations in Canvas.

I understand that maintaining backwards compatability with internals is uncommon for forks to do, but given that Canvas has worked in the past to maintain compatability with Folia plugins as much as possible, I figure that this PR could at the very least have a discussion.

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/world/level/chunk/ChunkAccess.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/world/level/chunk/ChunkAccess.java.patch
@@ -1,6 +1,6 @@
 --- a/net/minecraft/world/level/chunk/ChunkAccess.java
 +++ b/net/minecraft/world/level/chunk/ChunkAccess.java
-@@ -73,10 +_,170 @@
+@@ -73,10 +_,273 @@
      protected final @Nullable BlendingData blendingData;
      public final Map<Heightmap.Types, Heightmap> heightmaps = Maps.newEnumMap(Heightmap.Types.class);
      // Paper - rewrite chunk system
@@ -16,7 +16,7 @@
 +      When the chunk has a lot of block entities, this is much faster
 +      to use than the previous 'Map<BlockPos, BlockEntity> blockEntities'
 +     */
-+    // public final Map<BlockPos, BlockEntity> blockEntities = new Object2ObjectOpenHashMap<>();
++    public final Map<BlockPos, BlockEntity> blockEntities = new CanvasBlockEntityMapView();
 +    private static final BlockEntity[] EMPTY_ARRAY = new BlockEntity[0];
 +    private static final int SECTION_SIZE = 16 * 16;
 +    protected final BlockEntity[][] canvas$blockEntitySections;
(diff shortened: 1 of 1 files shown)


Dueris (approved):
I am completely for this, LGTM.

One note though, while not required, is I would like this map class to be placed under the `canvas$buildBlockEntityMap` function, to try and keep those set of respective changes in 1 clump together. I am not against it being at the bottom of the class, and while I normally would do the same, I feel that given we are in a patch-based environment, that the respective changes should be clumped together as much as possible to avoid excessive diff.

Beyond that, I agree with you and am all for this. Nice work

MrMasrozYTLIVE (approved):
LGTM

Jsinco:
Moved the wrapper up into the canvas changes block and cleaned up the diff a bit

R00tB33rMan (approved):

canvas-server/minecraft-patches/sources/net/minecraft/world/level/chunk/ChunkAccess.java.patch
```diff
@@ -171,6 +171,68 @@
...
++        public Set<Entry<BlockPos, BlockEntity>> entrySet() {
++            BlockEntity[] all = canvas$getAllBlockEntities();
++            Set<Entry<BlockPos, BlockEntity>> set = Sets.newLinkedHashSetWithExpectedSize(all.length);
++            for (final BlockEntity blockEntity : all) {
++                set.add(new java.util.AbstractMap.SimpleImmutableEntry<>(blockEntity.getBlockPos(), blockEntity));
++            }
++            return set;
++        }
++    }
++    // Canvas end - optimized collection: retain folia internals
```
Dueris: You do not need to add the `// Canvas end` here

canvas-server/minecraft-patches/sources/net/minecraft/world/level/chunk/ChunkAccess.java.patch
```diff
@@ -171,6 +171,68 @@
 +        });
 +        return this.canvas$cachedBlockEntityMap = java.util.Collections.unmodifiableMap(blockEntities);
 +    }
++    // Canvas start - optimized collection: retain folia internals
```
Dueris: You do not need to add the `// Canvas start` here

Dueris (changes requested):

Dueris:
Just a quick nitpick. Essentially, since this is already in the Canvas diff for the optimization in question and this is related to said optimization, there is no need to have those 2 comments for the changes made in this PR.

Jsinco:
Removed both start / end comments, left the `\n` breaking up space between the class def and other methods

canvas-server/minecraft-patches/sources/net/minecraft/world/level/chunk/ChunkAccess.java.patch
```diff
@@ -16,7 +16,7 @@
 +      When the chunk has a lot of block entities, this is much faster
 +      to use than the previous 'Map<BlockPos, BlockEntity> blockEntities'
 +     */
-+    // public final Map<BlockPos, BlockEntity> blockEntities = new Object2ObjectOpenHashMap<>();
++    public final Map<BlockPos, BlockEntity> blockEntities = new CanvasBlockEntityMapView(); // Canvas - optimized collection: retain folia internals
```
Dueris: Remove this comment too

Dueris (changes requested):

Dueris:
TBH I thought I had mentioned that, my bad

Jsinco:
no worries, it's gone now

Dueris (approved):

R00tB33rMan (approved):
Assumed that comment that was mentioned before was intended. LGTM again

MrMasrozYTLIVE (approved):
Still LGTM

devnootnoot:
Looks good to me overall, but there is one compatibility issue I think is worth fixing: `entrySet()` currently returns a newly created snapshot set containing `SimpleImmutableEntry` instances. That means the normal collection views inherited from `AbstractMap` are not actually backed by the underlying block entity storage.

For example, calls like these may only modify the temporary snapshot and leave the actual chunk data unchanged:

`chunk.blockEntities.entrySet().clear();`
`chunk.blockEntities.entrySet().remove(entry);`
`chunk.blockEntities.keySet().remove(pos);`
`chunk.blockEntities.values().remove(blockEntity);`
`chunk.blockEntities.entrySet().removeIf(...);`

Some of those operations may even return `true`, making it look like the block entity was removed when it was not. `Map.Entry#setValue` will also throw because the entries are immutable, whereas entries from the previous mutable map could normally write through.

Since the goal of this PR is Folia internals compatibility, I think `entrySet()` should return a proper backed view whose iterator/remove operations delegate to `canvas$removeFromBuckets`, and whose entries either support `setValue` through the bucket setter or explicitly document that mutation through entries is unsupported.

There is also a smaller map-contract edge case in `put(key, value)`: storage uses the supplied key coordinates, but iteration reconstructs the key using `value.getBlockPos()`. If those ever differ, `get(key)` and `entrySet()` disagree about what key the value belongs to. Internal callers probably always pass matching values, but it may be worth enforcing `key.equals(value.getBlockPos())` or building entries from the actual bucket coordinates.

Other than the collection-view behavior, the direct `get`, `containsKey`, `put`, `remove`, `clear`, and `iteration` paths look straightforward. The backed `entrySet` issue is the main thing I would address before calling this fully compatible with the old field.

Jsinco:
Wrapped. I decided not to add another static class for the sake of diff for right now, @Dueris or another maintainer can comment if they would prefer another static class or not to save on indents.

Toffikk (approved):

devnootnoot:
Unless I'm blind, the main issue I still see is that `entrySet()` returns a newly created snapshot made of `SimpleImmutableEntry` objects instead of a live view backed by the actual chunk storage.

Jsinco:
Are you sure? I don't see `SimpleImmutableEntry` used anywhere

devnootnoot:
ah yeah looked at the wrong thing. everything looks good now, sorry about that

devnootnoot (approved):

Dueris:
I pulled it into the post merger branch, `main`, in this commit: https://github.com/CraftCanvasMC/Canvas/commit/5bab338c3f73c7da560c63f7d1c7a15cccc92579

Thank you for your contribution, Im just merging it directly into there since the `pre-merger/26.2` branch wont be run by the CI again due to us merging Folia

PR #264, opened 2026-06-19
Improve command execution in region threading

Dueris:
## Overview

Both Folia and Canvas are known for cutting corners when in terms of restoring Vanilla commands to work with region threading. This often leads to issues or minor bugs or assumptions being made in command execution that lead to oddities. For example:

- Folia restores the `clear` command, which clears the inventory contents of the specified target player entities. Normally the command feedback(in English) is `Removed %s item(s) from player %s` for singular targets, and `Removed %s item(s) from %s players` for multiple targets. Folia breaks this feedback by always returning the first argument(the amount of items removed), as the same amount of entities in the target selection. This means, in Folia, if I had my inventory full of items stacked to 64, and I run the `clear` command on only myself, instead of returning `2304` items, Folia would provide feedback of `1` item being cleared. While minor, this sort of issue is present in numerous kinds of commands throughout Folia's patches.

There are also numerous cases where there is frankly, improper threading alltogether from commands restored by Folia. For example, the `swing` command is enabled in Folia while not doing any scheduling of any kind to the target(s) in question, which can lead to race conditions if the target is in a different region, given the `LivingEntity#swing` function is not thread-safe.

While these issues are minor and probably not going to be encountered all that often, these issues are yet *still present* and plentiful. The changes made in this PR aim to fix all of these issues found in the restored Vanilla commands by Canvas and Folia, since there simply lacked a system required to make these commands more closely aligned before. Both Canvas and Folia have altered functionality, and altered feedback responses, and more, to try and make commands work, but this PR brings these commands a bit closer to Vanilla, and fixes these issues, without an ugly af diff.

## The ACE API

The API in question is called the `AbstractCommandExecution` API(referred to later in this PR as "ACE"). The ACE API essentially allows defining targets, an executor, and a completion callback with abstract responses defined by the type `R` and `E extends Entity`. `R` defines the type of return the execution requires for the callback, and `E` defines the type of entity in question, like for if players specifically are required.

When constructing a new ACE instance, you are required to specify a singular `E` instance, a `Stream` of `E` instances, or a `Collection` of `E` instances. These define the entity targets that will be iterated over and scheduled to. You are also required to provide an executor, which is a functional interface that defines the logic for what happens on each entity in the target list and then what to do with the `R` data. The `R` data has synchronized mutation, meaning that when you provide the function to modify the `R` data, it will be held under locks, so no concurrency handling is required if working with a data type that is not thread-safe. The executor provides an `E` target instance, and is guaranteed that the code executed in the executor is on the owning thread of the target instance provided. An example(the `damage` command) is shown below:

```java
executes((Entity target) -> (_) -> target.hurtServer(stack.getLevel(), source, amount))
```

As you can see in the example, the executor requires returning a function that mutates the `R` data, since the `R` data is the data returned to the callback. The `hurtServer` function returns a `boolean` whether it failed to damage the entity or not, which is handled in the completion callback. The completion callback is provided the final `R` value, after being mutated across all entities in the target list, and is provided the `CommandSourceStack`, and the target list. It is worth noting that the completion callback has no guarantee of being on the owning context of the command source, so there should be no mutations to any entities at all at this point, and it should just be returning feedback of the command. The `damage` command completion callback is shown below:

```java
onComplete((passed, targets, css) -> {
    if (!passed) {
        throw ERROR_INVULNERABLE.create();
    }
    css.sendSuccess(() -> Component.translatable("commands.damage.success", amount, targets.getFirst().getDisplayName()), true);
    return Command.SINGLE_SUCCESS;
})
```

As you can see, the first variable passed is the `R` value, whether it passed or not. The second variable passed is the target list, and the third argument is the command source instance. In the `damage` command, note that the completion only checks the passed value and returns the feedback of the command *only*. It is also worth noting that the `return Command.SINGLE_SUCCESS` is simply there to simplify diff, since often times this API wraps existing Vanilla code, so we require returning an integer in response to mimic the same method descriptor as the previous enclosing method(since all commands require an integer return value). The full `damage` command reimplementation is shown below:

```java
io.canvasmc.canvas.command.execution.AbstractCommandExecution._boolean(false)
     .with(selected)
     .executes((Entity target) -> (_) -> target.hurtServer(stack.getLevel(), source, amount))
     .onComplete((passed, targets, css) -> {
         if (!passed) {
             throw ERROR_INVULNERABLE.create();
         }
         css.sendSuccess(() -> Component.translatable("commands.damage.success", amount, targets.getFirst().getDisplayName()), true);
         return Command.SINGLE_SUCCESS;
     })
     .start(stack);
```

Note, the `damage` command takes a singular target instance.

As you can see, the `damage` command takes a target, `selected`, and on the owning context of the target, it executes the `hurtServer` function and applies a mutation to the `R` value, setting it to the return value of the function, then passing it into the completion callback. The completion callback is always executed on the thread/region of the final executing entity in the target set. Ordering is not guaranteed, so no mutations should be done in the completion callback context.

Note the `start` function, which verifies the ACE definition and schedules everything, takes a `CommandSourceStack` instance, which is used to handle failures, exceptions, and be passed into the completion callback.

This system allows us to more closely align with Vanilla mechanics and behavior and more. This system also allowed this PR to fix a large sum of issues found in Folia and Canvas' restorations of certain commands.

## For reviewers

You may have to compare the diff in the source against the Folia source code and Paper source code, since some commands were fully reverted to the Paper equivalent before being re-regionized with this system.

## Extras

It was found during the creation of this PR that the stopwatches system is not properly regionized in Folia, and was recklessly enabled by Folia in [25w41a](https://minecraft.wiki/w/Java_Edition_25w41a), or `1.21.11` with no actual changes to this system to make it safe. This PR makes this system run very similarly to the scoreboards rewrite, requiring that all mutations are made on the **global tick**, while guaranteeing that all read operations are safe from any thread. Alongside this, autosave of this global data was redone to be a lot safer and better, only autosaving specific types of saved data, rather than **all** saved data, which may or may not be safe to save async.

## Side notes

All commands need extreme testing before merging. I did testing on my own to make sure they worked, however I maybe missed something or didn't test all the way. The modified commands should be tested in-game from multiple region contexts and in many scenarios to try and ensure that the changes made in this PR are safe for production.

### DIFF
--- a/canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
+++ b/canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
@@ -1843,7 +1843,7 @@ index fb55ed68a1f8470b551030a27ca2656555c93d9e..5756f7aca1a80782a88b0e82ef82a626
      public @Nullable CustomBossEvent get(final Identifier id) {
          return this.events.get(id);
 diff --git a/net/minecraft/server/commands/AdvancementCommands.java b/net/minecraft/server/commands/AdvancementCommands.java
-index dfabf9b9870091087ab8b48634ecf9e2316b7f68..c72e69136b591947d0901d9f64b353706262520c 100644
+index dfabf9b9870091087ab8b48634ecf9e2316b7f68..ffdf3e49d9366e89940a043907b2d83ab62b22ba 100644
 --- a/net/minecraft/server/commands/AdvancementCommands.java
 +++ b/net/minecraft/server/commands/AdvancementCommands.java
 @@ -238,23 +238,38 @@ public class AdvancementCommands {
@@ -1874,8 +1874,8 @@ index dfabf9b9870091087ab8b48634ecf9e2316b7f68..c72e69136b591947d0901d9f64b35370
 +            .executes((ServerPlayer entityPlayer) -> {
 +                int changedAdvancements = action.perform(entityPlayer, advancements, showAdvancements);
 +                return io.canvasmc.canvas.command.execution.ExecutionCallbacks.mapPairBoth(
-+                    (ac) -> ac + changedAdvancements,
-+                    (pc) -> changedAdvancements > 0 ? pc + 1 : pc
++                    (advancementCount) -> advancementCount + changedAdvancements,
++                    (playerCount) -> changedAdvancements > 0 ? playerCount + 1 : playerCount
 +                );
 +            })
 +            .onComplete((pair, _, css) -> performCallback(
--- a/canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
+++ b/canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
@@ -2271,7 +2271,7 @@ index 1aead2cd0b8fa8e61ba4820ec116ec9e0a5c0bfe..2cca48a398f296bc193d35994ffcae4b
  
      private static Component getAttributeDescription(final Holder<Attribute> attribute) {
 diff --git a/net/minecraft/server/commands/BossBarCommands.java b/net/minecraft/server/commands/BossBarCommands.java
-index 5cbe5da7013faadc7d0e9eb03ff5738dc1afab17..3de09a5e689049cbb94268d257556bc2c7af39f9 100644
+index 5cbe5da7013faadc7d0e9eb03ff5738dc1afab17..89c7ca62c01d43e6899a57ee978e7a121d11140e 100644
 --- a/net/minecraft/server/commands/BossBarCommands.java
 +++ b/net/minecraft/server/commands/BossBarCommands.java
 @@ -176,16 +176,21 @@ public class BossBarCommands {
@@ -2416,7 +2416,7 @@ index 5cbe5da7013faadc7d0e9eb03ff5738dc1afab17..3de09a5e689049cbb94268d257556bc2
          CustomBossEvents events = source.getServer().getCustomBossEvents();
          if (events.get(id) != null) {
              throw ERROR_ALREADY_EXISTS.create(id.toString());
-@@ -333,17 +358,21 @@ public class BossBarCommands {
+@@ -333,14 +358,17 @@ public class BossBarCommands {
          );
          source.sendSuccess(() -> Component.translatable("commands.bossbar.create.success", event.getDisplayName()), true);
          return events.getEvents().size();
@@ -2434,10 +2434,6 @@ index 5cbe5da7013faadc7d0e9eb03ff5738dc1afab17..3de09a5e689049cbb94268d257556bc2
      }
  
      public static CustomBossEvent getBossBar(final CommandContext<CommandSourceStack> context) throws CommandSyntaxException {
-+        io.papermc.paper.threadedregions.RegionizedServer.ensureGlobalTickThread("Cannot get bossbar async"); // Canvas - region threading
-         Identifier id = IdentifierArgument.getId(context, "id");
-         CustomBossEvent event = context.getSource().getServer().getCustomBossEvents().get(id);
-         if (event == null) {
 diff --git a/net/minecraft/server/commands/ClearInventoryCommands.java b/net/minecraft/server/commands/ClearInventoryCommands.java
 index dc68446fa7959848ad8d6497b8f5118d55e910a6..d5ddf780a45daaf27bdaf72b1bf20c35634e4c37 100644
 --- a/net/minecraft/server/commands/ClearInventoryCommands.java
--- a/canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
+++ b/canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
@@ -212,7 +212,7 @@ index 0ce4643d526368b85ead19dd5d63aa41ec41afc6..8834c7fb6edff6c0a02231dd61e01620
          LOGGER.info("Halting chunk systems...");
          for (final ServerLevel world : MinecraftServer.getServer().getAllLevels()) {
 diff --git a/io/papermc/paper/threadedregions/RegionizedServer.java b/io/papermc/paper/threadedregions/RegionizedServer.java
-index 918d58644ac028a1025811ebd459112ff6a0e78f..be02125b9f01ee60df771f83aa91551f54755f57 100644
+index 918d58644ac028a1025811ebd459112ff6a0e78f..54baa6300da3209c912a0405e23492c86cf28254 100644
 --- a/io/papermc/paper/threadedregions/RegionizedServer.java
 +++ b/io/papermc/paper/threadedregions/RegionizedServer.java
 @@ -31,7 +31,7 @@ import java.util.function.BooleanSupplier;
@@ -272,11 +272,11 @@ index 918d58644ac028a1025811ebd459112ff6a0e78f..be02125b9f01ee60df771f83aa91551f
          // connection tick
 +        // ender pearls
 +        // gui tick
-+        // scoreboard
++        // scoreboard, stopwatches
  
          // tick player ping sample
          this.tickPlayerSample();
-@@ -362,6 +377,17 @@ public final class RegionizedServer {
+@@ -362,7 +377,63 @@ public final class RegionizedServer {
  
          // player list
          MinecraftServer.getServer().getPlayerList().tick();
@@ -288,13 +288,59 @@ index 918d58644ac028a1025811ebd459112ff6a0e78f..be02125b9f01ee60df771f83aa91551f
 +        // gui tick
 +        net.minecraft.server.gui.PlayerListComponent.tick();
 +
-+        // scoreboard
-+        MinecraftServer.getServer().getScoreboard().autosave();
++        // scoreboards, stopwatches
++        autosaveSafeGlobalData(false, false);
 +        // Canvas end - region threading
++    }
++    // Canvas start - region threading
++
++    private static final java.util.concurrent.atomic.AtomicLong LAST_GLOBAL_AUTOSAVE = new java.util.concurrent.atomic.AtomicLong(System.nanoTime());
++
++    // note: not everything is thread-safe, only a few things are thread-safe
++    //       due to specific design of these reimplementations. those are:
++    //  - scoreboards: from Canvas scoreboards rewrite
++    //  - stopwatches: we made this SWMR fully, and only the global tick can
++    //                 write to the backend map so this is safe to save here
++    // note: this does not save maps at all, those are per-world
++    public void autosaveSafeGlobalData(final boolean force, final boolean sync) {
++        ensureGlobalTickThread("Cannot save global data async");
++
++        final MinecraftServer server = MinecraftServer.getServer();
++
++        final int autoSavePeriod = server.autosavePeriod;
++        if (autoSavePeriod <= 0 && !force) return;
++        final long periodNanos = autoSavePeriod * io.papermc.paper.threadedregions.TickRegionScheduler.getTimeBetweenTicks();
++        final long currTime = System.nanoTime();
++        final long last = LAST_GLOBAL_AUTOSAVE.get();
++
++        if (force || ((currTime - last) > periodNanos && LAST_GLOBAL_AUTOSAVE.compareAndSet(last, currTime))) {
++
++            // these 2 are the only safe global data ones besides maps,
++            // however those are per-world and saved in world global tick
++
++            // we have to do this for the scoreboard system or else the scoreboard
++            // system won't actually save. see MinecraftServer#saveGlobalData
++
++            server.getScoreboard().storeToSaveDataIfDirty(server.getDataStorage().computeIfAbsent(net.minecraft.world.scores.ScoreboardSaveData.TYPE));
++
++            java.util.concurrent.CompletableFuture<?> future = server.getDataStorage().saveSpecific(
++                net.minecraft.world.scores.ScoreboardSaveData.class,
++                net.minecraft.world.Stopwatches.class
++            );
++
++            // if sync we need to just join here. this shouldn't be done
++            // for autosave or every tick since this blocks waiting for completion
++
++            if (sync) {
++                future.join();
++            }
++        }
      }
++    // Canvas end - region threading
  
      private void tickPlayerSample() {
-@@ -402,9 +428,14 @@ public final class RegionizedServer {
+         final MinecraftServer mcServer = MinecraftServer.getServer();
+@@ -402,9 +473,14 @@ public final class RegionizedServer {
              }
  
              if (isNotOwnedByGlobalRegion(conn)) {
@@ -312,7 +358,7 @@ index 918d58644ac028a1025811ebd459112ff6a0e78f..be02125b9f01ee60df771f83aa91551f
              }
  
              if (!conn.isConnected()) {
-@@ -435,6 +466,12 @@ public final class RegionizedServer {
+@@ -435,6 +511,12 @@ public final class RegionizedServer {
      // A global tick only updates things like weather / worldborder, basically anything in the world that is
      // NOT tied to a specific region, but rather shared amongst all of them.
      private void globalTick(final ServerLevel world, final long tickCount) {
@@ -325,7 +371,7 @@ index 918d58644ac028a1025811ebd459112ff6a0e78f..be02125b9f01ee60df771f83aa91551f
          // needs
          // worldborder tick
          // advancing the weather cycle
-@@ -480,6 +517,7 @@ public final class RegionizedServer {
+@@ -480,6 +562,7 @@ public final class RegionizedServer {
      }
  
      private void advanceWeatherCycle(final ServerLevel world) {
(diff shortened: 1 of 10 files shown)


yunuservices (dismissed):
There are two blockers I would want addressed first on my perspective

1. `AbstractCommandExecution` currently hardcodes command execution helpers to return `Command.SINGLE_SUCCESS` instead of propagating the real `action.act()` result. That changes vanilla command result semantics for every caller using this abstraction, including `execute store result` and any count/value-returning command path.

2. `team empty` has been removed from the dispatcher. That is a user-visible feature regression. If this path is not safe yet, it should be fixed or deferred separately.

canvas-server/src/main/java/io/canvasmc/canvas/command/execution/AbstractCommandExecution.java
```diff
@@ -0,0 +1,272 @@
...
+        level.canvas$loadOrRunAtChunksAsync(
+            pos, radius, Priority.NORMAL, () -> {
+                try {
+                    action.act();
+                } catch (final CommandSyntaxException cse) {
+                    css.sendFailure(Component.literal(cse.getMessage()));
+                }
+            }
+        );
+        return Command.SINGLE_SUCCESS;
```
yunuservices, line 82: This hardcodes the command result to `SINGLE_SUCCESS` and drops the actual `action.act()` return value.

That changes vanilla command result semantics for every caller using this helper, including `execute store result` and count/value-returning command paths.

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
yunuservices, line 1: Removing `team empty` from the dispatcher is a user-visible feature regression.

If this path is not safe yet, it should be fixed or deferred separately, but it should not be silently removed in this PR.

canvas-server/src/main/java/io/canvasmc/canvas/command/execution/AbstractCommandExecution.java
Dueris: Canvas and Folia do not implement the command integer return value. We never kept this parity and keeping that parity is impossible.

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
Dueris: It wasnt removed in this PR, we already removed it. Currently I dont have an implementation ready for deployment to reintroduce this

Dueris:
Note that before this is merged, docs need to be written on this

I can handle that, Ill write it once this PR is approved and before it's merged. Ill merge this PR when everyone approves and Ive handled documentation

devnootnoot:
Coordinate conversion in `executeAtPos` / `executeAtPosWithRadius` is a bit weird. The `Vec3` overloads use casts like `(int) pos.x` and `(int) pos.z`. Java casts truncate toward zero, so negative fractional coordinates resolve incorrectly. For example, `-0.5` becomes `0`, while the containing block should be `-1`. In `executeAtPos`, that can also select the wrong chunk and therefore potentially schedule the command onto the wrong region. I think these should use `BlockPos.containing(pos)` or `Mth.floor(...)` instead.

The exception flow in `AbstractCommandExecution` also seems a little awkward. If one target throws a `CommandSyntaxException`, a failure is sent immediately, but once the final target finishes, the normal completion callback still runs. Depending on the command, that callback may then send a success message using the partially collected result, so the source could receive both a failure and a success for the same command. It might be worth tracking whether any execution failed and either skipping the normal success callback or giving the completion callback the failure state explicitly.

Related to that, the execution count is only decremented for successful executions or caught `CommandSyntaxExceptions`. If the command action or data modifier throws an unchecked exception, the count is never decremented and the completion callback never runs. The scheduler may log the exception, but the ACE instance stays permanently incomplete from its own point of view. A finally-based decrement, with proper exception reporting, would make the lifecycle more robust.

One smaller typo: the validation errors for a missing command action and missing targets both currently report `"on complete"`, which will make configuration mistakes confusing to debug.

Correct me if I'm wrong on any of these points, this is just from a quick glance. I do like the PR and think it's great overall.

devnootnoot:
LGTM

devnootnoot (approved):

Jsinco (approved):
Ran through and tested the commands modified in this PR in a single region and also across multiple regions and everything works as expected. The one concern I had has already been addressed (over discord) in cf42c4e so everything else looks good to me.

Other than that, I believe that in a future PR after this is merged this framework should be exposed in the API.

Overall, nice work!

coredex-source (approved):
Tested the lastest commits as well in PROD. LGTM and works really well!

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
```diff
@@ -1812,8 +1842,639 @@ index fb55ed68a1f8470b551030a27ca2656555c93d9e..5756f7aca1a80782a88b0e82ef82a626
...
+-            }
+-
+-            advancementCount += changedAdvancements;
+-        }
++    // Canvas start - region threading
++        io.canvasmc.canvas.command.execution.AbstractCommandExecution.
++                <io.canvasmc.canvas.command.execution.ExecutionCallbacks.NumericPair<Integer, Integer>, ServerPlayer>
++                _abstract(new io.canvasmc.canvas.command.execution.ExecutionCallbacks.NumericPair<>(0, 0))
++            .with(players)
++            .executes((ServerPlayer entityPlayer) -> {
```
Toffikk, line 1874: change this to sourcePlayer

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
```diff
@@ -1812,8 +1842,639 @@ index fb55ed68a1f8470b551030a27ca2656555c93d9e..5756f7aca1a80782a88b0e82ef82a626
...
++                int changedAdvancements = action.perform(entityPlayer, advancements, showAdvancements);
++                return io.canvasmc.canvas.command.execution.ExecutionCallbacks.mapPairBoth(
++                    (ac) -> ac + changedAdvancements,
```
Toffikk: this could use a better name than ac and pc

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
```diff
@@ -1812,8 +1842,639 @@ index fb55ed68a1f8470b551030a27ca2656555c93d9e..5756f7aca1a80782a88b0e82ef82a626
...
+-        int playerCount = 0;
++        // Canvas - region threading
+         Advancement advancement = holder.value();
+         if (!advancement.criteria().containsKey(criterion)) {
+             throw ERROR_CRITERION_NOT_FOUND.create(Advancement.name(holder), criterion);
+         }
++    // Canvas start - region threading
++        io.canvasmc.canvas.command.execution.AbstractCommandExecution.<ServerPlayer>_int()
++            .with(players)
++            .executes((ServerPlayer entityPlayer) -> {
```
Toffikk, line 1917: sourcePlayer

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
```diff
@@ -1812,8 +1842,639 @@ index fb55ed68a1f8470b551030a27ca2656555c93d9e..5756f7aca1a80782a88b0e82ef82a626
...
+-                sendMessage(source, ex);
+-            }
+-        });
+-        return 0;
+-        // Folia end - region threading
++        // Canvas start - region threading
++        io.canvasmc.canvas.command.execution.AbstractCommandExecution._double()
++            .with(target)
++            .executes((Entity entity) -> {
++                LivingEntity livingEntity = getEntityWithAttribute(entity, attribute);
```
Toffikk, line 1994: maybe it’d better to change this to `living` as is the case in some other classes iirc

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
```diff
@@ -2056,69 +2725,881 @@ index 50d671bc5a8860883239590c030881043c791ca1..92f1416b3994e7f1334bdf41f8c00f61
...
+     }
+ 
+-    // Folia start - region threading
+-    private static void sendMessage(CommandSourceStack src, CommandSyntaxException ex) {
+-        src.sendFailure((Component)ex.getRawMessage());
+-    }
+-    // Folia end - region threading
+-
+     private static int enchant(
+         final CommandSourceStack source, final Collection<? extends Entity> targets, final Holder<Enchantment> enchantmentHolder, final int level
```
Toffikk, line 3030: enchantment

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
```diff
@@ -2056,69 +2725,881 @@ index 50d671bc5a8860883239590c030881043c791ca1..92f1416b3994e7f1334bdf41f8c00f61
...
++            })
++            .onComplete((changed, _, css) -> enchantCallback(css, changed, enchantmentHolder, level))
++            .start(source);
++        return 1;
++    }
++
++    private static int enchantCallback(
++        final CommandSourceStack source,
++        final io.canvasmc.canvas.util.CollectingList<Entity> targets,
++        final Holder<Enchantment> enchantmentHolder,
```
Toffikk, line 3107: here too

Toffikk (commented):
some of these comments apply to other places but i just didnt feel like commenting everywhere. i havent reviewed everything yet

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
Dueris: It should not. The `executes` function runs over all the players in the selection iteration. There is no guarantee which entity is the source or even if the source is in the selection, so `sourcePlayer` wouldn't make sense

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
Dueris: See my previous response

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
Dueris: This is Vanilla diff, and should not be changed

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
Dueris: That is copied from the Vanilla method params of the original unified function. This should remain as-is

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
Dueris: This is what it was in the Vanilla diff, this has no reason to be changed since this is a copy from the original function

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
Dueris: This has been changed to match their original Vanilla var names

R00tB33rMan (approved):
Got a test conducted on a sep. instance and the latest iteration looks solid!

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
```diff
@@ -1812,8 +1842,639 @@ index fb55ed68a1f8470b551030a27ca2656555c93d9e..5756f7aca1a80782a88b0e82ef82a626
...
+-                action.perform(newPlayer, advancements, showAdvancements);
+-            });
+-            // Folia end - region threading
+-            if (changedAdvancements > 0) {
+-                playerCount++;
+-            }
+-
+-            advancementCount += changedAdvancements;
+-        }
++    // Canvas start - region threading
```
BaconCat1, line 1869: Consider fixing indentation here for `// Canvas` comment.

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
```diff
@@ -1343,7 +1389,7 @@ index 585dcf881008e28aba077092943d4cbf4eb5f5b4..1307beab98cd028ac8087aa32aa499fb
 -        //TagCommand.register(this.dispatcher); // Folia - region threading - TODO later
 -        //TeamCommand.register(this.dispatcher, context); // Folia - region threading - TODO later
 -        //TeamMsgCommand.register(this.dispatcher); // Folia - region threading - TODO later
-+        TagCommand.register(this.dispatcher); // Canvas - TODO - do better
++        TagCommand.register(this.dispatcher);
```
BaconCat1, line 1392: Missing `// Canvas` comment.

canvas-server/src/main/java/io/canvasmc/canvas/command/execution/AbstractCommandExecution.java
```diff
@@ -0,0 +1,290 @@
...
+                if (uncaught instanceof CommandSyntaxException cse) {
+                    sourceStack.sendFailure(Component.literal(cse.getMessage()));
+                }
+
+                // realistically, this should kill the server if not a CSE...in theory
+                // but only if this was scheduled. if it wasn't this will do nothing so we
+                // need to set "has failed" anyway or else this may continue
+
+                hasFailed.set(true);
+                throw new RuntimeException("Uncaught exception when executing ACE constructed command", uncaught);
```
BaconCat1, line 283: When `uncaught` is a `CommandSyntaxException`, we send it to the source but are still rethrow it as a `runtimeException`. Should normal command validation errors stop the ACE flow?

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
```diff
@@ -2403,350 +3862,820 @@ index 57e8501f4adc66a11a474905bed045c34a44c9cb..a920e8594832e99acfe1cce667f9fe37
...
++            })
++            .onComplete((result, processed, css) -> performOperationCallback(result, processed.stream().map(ScoreHolder.class::cast).toList(), targetObjective, css))
++            .start(source);
++        return 1;
++    }
++
++    private static final com.mojang.brigadier.exceptions.DynamicCommandExceptionType TARGET_MUST_BE_ENTITY = new com.mojang.brigadier.exceptions.DynamicCommandExceptionType(
++        name -> Component.literal("Holder target \"" + name + "\" must be an entity")
++    );
++    private static final java.util.function.Function<net.minecraft.world.scores.ScoreHolder, net.minecraft.world.entity.Entity> SCORE_HOLDER2ENTITY = (name) -> {
```
BaconCat1, line 4405: Feel like I recall some plugins using fake score holders that aren't entities. Should we consider supporting this or is there some reason `ScoreHolder` must always be an entity?

canvas-server/src/main/java/io/canvasmc/canvas/command/execution/AbstractCommandExecution.java
```diff
@@ -0,0 +1,290 @@
...
+            return;
+        }
+
+        // iterate over all targets now and schedule or execute
+        for (final E selected : targets) {
+            if (TickThread.isTickThreadFor(selected)) {
+                action.accept(selected);
+            }
+            else {
+                selected.getBukkitEntity().taskScheduler.schedule(
```
BaconCat1: The return value of this is ignored here. If scheduling fails count may never reach zero and completion may never fire.

BaconCat1 (dismissed):

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
Dueris: This is intentional. The tag command was improved in this PR so the comment is no longer needed

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
Dueris: Intentional. See the end comment below. Canvas comments always align the start and end comments

canvas-server/src/main/java/io/canvasmc/canvas/command/execution/AbstractCommandExecution.java
Dueris: Fixed

canvas-server/minecraft-patches/base/0004-Fixup-Region-Threading.patch
Dueris: Plugins are able to use the scoreboard API to do so. This is strictly only for commands. This is for scheduling purposes specifically.

The Bukkit API still supports fake scoreholders

canvas-server/src/main/java/io/canvasmc/canvas/command/execution/AbstractCommandExecution.java
Dueris: Fixed

Toffikk (approved):
lgtm code-wise

PR #265, opened 2026-06-20
Fix MC-261810

Dueris:
https://bugs.mojang.com/browse/MC/issues/MC-261810

When in the void, rockets get discarded during the rocket entity tick, thus making it so that rockets are practically ineffective if in the void with an elytra(in the void meaning you are taking damage).

This fixes it by not allowing the rocket entity to be discarded when in the void if attached to an entity in the same region.

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/projectile/FireworkRocketEntity.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/projectile/FireworkRocketEntity.java.patch
@@ -1,5 +1,29 @@
 --- a/net/minecraft/world/entity/projectile/FireworkRocketEntity.java
 +++ b/net/minecraft/world/entity/projectile/FireworkRocketEntity.java
+@@ -126,6 +_,23 @@
+         super.inactiveTick();
+     }
+     // Paper end - EAR 2
++    // Canvas start - MC-261810
++
++    @Override
++    protected void onBelowWorld() {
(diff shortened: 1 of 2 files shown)


Dueris:
While out of scope of this PR, we should restore the comments on these config options...

R00tB33rMan (approved):

devnootnoot:
The only thing I would maybe consider is whether the condition should also check that the attached entity is still alive/not removed. I do not think this is a major issue because the rocket still has its normal lifetime handling, but if the attached entity can become removed while the rocket still keeps a reference, this could allow the rocket to avoid below-world discard until its lifetime expires.

Dueris:
In Vanilla from my testing, it seems that the logic is that the firework entity will still apply momentum to the attached target in question upon respawn.

To try, enable keep inventory and immediate respawn and then in creative spam fireworks while gliding and then run `/kill` to observe this behavior. Ensure you are next to your spawn location though so it's in the same tick region.

I think the logic should be kept as-is to retain this functionality, since if we check on remove and say, those circumstances are met, then we would kinda more-or-less break Vanilla here. While a stupid scenario to actually run into, I feel that probably keeping it as-is is the way to go if we want Vanilla compat here for the most part

devnootnoot:
alright, if that's the goal then LGTM

devnootnoot (approved):

coredex-source (approved):

Toffikk (approved):

PR #266, opened 2026-06-22
Fix the Paper enablePlayerCollisions option

Dueris:
This option needs to be tested a bit more to be sure this works, I got someone that will hopefully test soon.

This removes our patches that disable this config option, since not only do we restore the system this relies on, but this works in Folia already, so this should be fine to enable functionally again.


R00tB33rMan (approved):
I personally haven’t seen bug reports nor issues with collisions when testing. Most production networks that encounter issues with player collisions tend to actually see more prevalent issues with plugins like TAB, where they disable collisions via Paper but keep them enabled via TAB, when they should instead keep them enabled in Paper and disable them in TAB to prevent further issues

Toffikk (approved):

Dueris:
Testing passed, works.

PR #269, opened 2026-06-23
Implement selective autosave

Dueris:
Amongst the numerous things Folia breaks, autosave is one of these things. In Vanilla Folia, it only autosaves chunks on unload, and maps. Thats it. Canvas already fixes pearls(with it's own autosave system), players(thanks to @Euphillya), scoreboards, and recently, stopwatches. While this is a step in the right direction, this PR intends to close this gap even further by doing the following:

- Rewrite our ender pearls system to use the `SavedData` system provided by Minecraft. This actually doesn't require any migration, given we mimicked the placement of this `dat` file already. This only required adjusting and removing our previous autosave system.
- Autosave clocks/time. This is already ticked on the global tick, which is where we conduct initial serialization before submitting writing to the IO utility pool. This was safe to do already, but we made a few adjustments to ensure that clocks are safe to autosave on the global tick throughout the server lifecycle by preventing modifications off the global tick(Folia already does this in some places), and wrapping the `clocks` map as an immutable map(already considered immutable prior to these changes, however this just enforces it).
- Autosave weather. This was already safe to do on the global tick, we just add 3 new thread checks to enforce this safety
- Autosave stopwatches and scoreboards. Already made thread-safe, just adjusted the impl triggering these so that they follow in-line with the rest of this data
- Autosave custom boss bars. Bossbars were made completely thread-safe in this PR. The codec that builds this data snapshots the current bossbars into an immutable copy of it before encoding.
- Autosave gamerules. This was already made thread-safe in Folia, just adjusted slightly to make it slightly more thread-safe and make the codec safe.
- Autosave dragon fight data. This is the only bit of data not autosaved on the global tick, and is instread autosaved at the end of the dragon fight tick in whatever region owns the dragon fight in question if the data is dirty. This is guarenteed to be 1 dragon fight per dimension(marked with a dragon fight in the dimension type definition) in 1 region.

Maps are already autosaved, so they are included in this.

It is worth noting the systems that are now autosaved are 1:1 with Vanilla since we directly restore the Vanilla system itself, excluding ender pearls ofc.

This way, we restore the Vanilla `SavedData` instances being autosaved on the Bukkit-defined interval if the `SavedData` in question is dirty, making this a lot safer than before, especially if unexpected crashes like a fatal JRE error or an OOM or something happens.


Euphillya (approved):

Dueris:
Any changes made to this branch at the result of reviews and stuff should be reflected in the `main` branch as this will not be squashed into the merger branch.

Toffikk (approved):

R00tB33rMan (approved):
LGTM - did testing on associated branches

devnootnoot:
In `autosaveSafeGlobalData`, the timestamp only gets updated through the `compareAndSet` inside the non-forced condition:

```
if (force || ((currTime - last) > periodNanos
    && LAST_GLOBAL_AUTOSAVE.compareAndSet(last, currTime))) {
```

Because `force` short-circuits the expression, a forced save such as `/save-all` does not update `LAST_GLOBAL_AUTOSAVE`. If the normal interval was already due, the next global tick can immediately trigger another autosave right after the forced synchronous save. I think the forced path should update the timestamp too.

`EnderPearls#isDirty()` also always returns `true`. That guarantees correctness, but it means the pearl file will be serialized and written on every global autosave even when nothing changed. Since this PR is specifically implementing selective dirty autosaving, it might be better to call `setDirty()` when pearls are added or removed and use the normal `SavedData` dirty flag.

The other pearl concern is in `spawnPearls`. Each stored pearl is removed from the saved list immediately after `Pearl#spawn()` is called, but the actual spawn happens later through `canvas$loadOrRunAtChunksAsync`. If the world is missing, entity decoding fails, chunk loading fails, or `tryAddFreshEntityWithPassengers` returns false, the saved pearl has already been removed and is permanently lost. I think the stored entry should only be removed after the async spawn succeeds, or restored/retained when spawning fails.

One smaller thing: the bossbar player update path appears to remove every player and add them all again, then always returns `true`. Even when the player set is unchanged, this can resend packets, briefly flicker the bar, and mark it dirty unnecessarily. Comparing the old/new sets or only applying the actual differences would avoid that.

Dueris:
This PR is already merged in the `main` branch. Fixups have been applied to address issues brought up here. Closing this as this has no point in being merged since this isn't going to target any specific relevant branch.

PR #272, opened 2026-06-28
Use concurrent map (accessed by multiple threads)

WouterGritter:
Use a `ConcurrentHashMap` as the `resources` field is accessed across multiple threads.

A CME is never thrown, and this map is only accessed through the atomic `computeIfAbsent`, meaning a concurrent operation would currently (theoretically) only be resulting in a duplicate compute of the value, which is benign. However this came up when looking into the following exception:

```
java.lang.NullPointerException: Cannot invoke "java.nio.file.Path.getFileSystem()" because "path" is null
    at java.base/java.nio.file.Files.provider(Files.java:103)
    at java.base/java.nio.file.Files.exists(Files.java:2289)
    at net.minecraft.util.FileUtil.createDirectoriesSafe(FileUtil.java:174)
    at net.minecraft.stats.ServerStatsCounter.save(ServerStatsCounter.java:94)
    at net.minecraft.server.players.PlayerList.save(PlayerList.java:486)
    at net.minecraft.server.players.PlayerList.saveAll(PlayerList.java:1125)
```

Where the NPE comes from a `null` `Path#getParent`, of a `Path` returned/computed by/in this map. I don't necessarily see how this could be resolved by this PR, but it's a worth while defensive change.

### DIFF
--- a/canvas-server/minecraft-patches/base/0004-Region-Threading.patch
+++ b/canvas-server/minecraft-patches/base/0004-Region-Threading.patch
@@ -29856,6 +29856,19 @@ index f05ab5fdab4cb587974e2864e859e49375f8165e..afecd66784ed57949b54032a1d6a5f44
          }
      }
  
+diff --git a/net/minecraft/world/level/storage/LevelStorageSource.java b/net/minecraft/world/level/storage/LevelStorageSource.java
+index 6f24eb34e0ecc2b03371c585a9538e9337a2295a..0dc70356eba5ea022e0991a81e11c9fe5812e926 100644
+--- a/net/minecraft/world/level/storage/LevelStorageSource.java
++++ b/net/minecraft/world/level/storage/LevelStorageSource.java
+@@ -460,7 +460,7 @@ public class LevelStorageSource {
+         private DirectoryLock lock;
+         private final LevelStorageSource.LevelDirectory levelDirectory;
+         private final String levelId;
+-        private final Map<LevelResource, Path> resources = Maps.newHashMap();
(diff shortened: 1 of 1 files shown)


Toffikk (approved):

R00tB33rMan (approved):

canvas-server/minecraft-patches/base/0004-Region-Threading.patch
```diff
@@ -29850,6 +29850,19 @@ index f05ab5fdab4cb587974e2864e859e49375f8165e..afecd66784ed57949b54032a1d6a5f44
...
+diff --git a/net/minecraft/world/level/storage/LevelStorageSource.java b/net/minecraft/world/level/storage/LevelStorageSource.java
+index 6f24eb34e0ecc2b03371c585a9538e9337a2295a..a538cc271af17d71992b65eac19bc0006ae50dae 100644
+--- a/net/minecraft/world/level/storage/LevelStorageSource.java
++++ b/net/minecraft/world/level/storage/LevelStorageSource.java
+@@ -460,7 +460,7 @@ public class LevelStorageSource {
+         private DirectoryLock lock;
+         private final LevelStorageSource.LevelDirectory levelDirectory;
+         private final String levelId;
+-        private final Map<LevelResource, Path> resources = Maps.newHashMap();
++        private final Map<LevelResource, Path> resources = Maps.newConcurrentMap(); // Canvas - concurrent map, accessed by multiple threads
```
Dueris: Can this comment become `// Canvas - region threading - accessed by multiple threads, use concurrent equivalent` or something? Tbh I only really want `- region threading` infront

Dueris (dismissed):

Dueris (approved):

PR #280, opened 2026-07-05
Update FasterRandomSource.java

ichbinrather:
the fromHashOf method was making an incorrect conversion when converting strings to random numbers the hash code of the string is a 32 bit number when converting this to 64 bits the upper part was always getting the same values because of this different strings were converting into the same number and everything was generating the same in world generation with the fix this issue was resolved


coredex-source:
The PR is incorrect and does not solve anything rather it makes things worse.

PR #281, opened 2026-07-06
Features: Each region could have its own weather.

Euphillya:
## Description
Enabling this setting allows each region on Folia to have its own weather, rather than a single global setting for everyone (disabled by default).
This could be ideal for Skyblock, OneBlock, and similar server types.

## Currently incompatible (for the time being)
Plugins that attempt to modify the weather, whether globally or otherwise.
WeatherControllerImpl is not called in API

## Additional feature
Added a `get` option to the `/weather` command to check the weather in the current region.

## Note
This PR is essentially a copy-paste from my own fork.

### DIFF
--- a/canvas-api/src/main/java/io/canvasmc/canvas/world/weather/WeatherController.java
+++ b/canvas-api/src/main/java/io/canvasmc/canvas/world/weather/WeatherController.java
@@ -2,23 +2,75 @@ package io.canvasmc.canvas.world.weather;
 
 import org.bukkit.Location;
 import org.bukkit.World;
+import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
+/**
+ * API to read and change the weather of individual ticking regions,
+ * when per-region weather is enabled.
+ *
+ * <p>Must be called from the tick thread owning the targeted region.</p>
+ */
+@NullMarked
 public interface WeatherController {
 
+    /**
+     * Sets the weather of the region owning the given chunk.
+     *
+     * @param world         the world containing the region
+     * @param chunkX        chunk X coordinate used to locate the region
+     * @param chunkZ        chunk Z coordinate used to locate the region
+     * @param type          target weather ({@code CLEAR}, {@code RAIN} or {@code THUNDER})
+     * @param durationTicks how long to hold this weather, in ticks; a value {@code <= 0}
+     *                      lets the current weather cycle continue naturally
+     * @param instantLevels if {@code true}, applies the rain/thunder intensities
+     *                      immediately instead of fading progressively
+     * @throws IllegalStateException if not called from the region's tick thread
+     */
     void setRegionalWeather(World world, int chunkX, int chunkZ, WeatherType type, int durationTicks, boolean instantLevels);
 
+    /**
+     * Sets the weather of the region owning the given location.
+     *
+     * @param loc           the location whose chunk determines the region
+     * @param type          target weather ({@code CLEAR}, {@code RAIN} or {@code THUNDER})
+     * @param durationTicks how long to hold this weather, in ticks; a value {@code <= 0}
+     *                      lets the current weather cycle continue naturally
+     * @param instantLevels if {@code true}, applies the rain/thunder intensities
+     *                      immediately instead of fading progressively
+     * @throws IllegalStateException if not called from the region's tick thread
+     * @see #setRegionalWeather(World, int, int, WeatherType, int, boolean)
+     */
     default void setRegionalWeather(Location loc, WeatherType type, int durationTicks, boolean instantLevels) {
-        final int cx = loc.getBlockX() >> 4;
-        final int cz = loc.getBlockZ() >> 4;
-        setRegionalWeather(loc.getWorld(), cx, cz, type, durationTicks, instantLevels);
+        final int chunkX = loc.getBlockX() >> 4;
+        final int chunkZ = loc.getBlockZ() >> 4;
+        setRegionalWeather(loc.getWorld(), chunkX, chunkZ, type, durationTicks, instantLevels);
     }
 
+    /**
+     * Returns a snapshot of the weather of the region owning the given chunk.
+     *
+     * @param world  the world containing the region
+     * @param chunkX chunk X coordinate used to locate the region
+     * @param chunkZ chunk Z coordinate used to locate the region
+     * @return the current weather state of the region, or {@code null} if
+     * per-region weather is disabled for this world
+     * @throws IllegalStateException if not called from the region's tick thread
+     */
     @Nullable RegionalWeatherSnapshot getRegionalWeather(World world, int chunkX, int chunkZ);
 
+    /**
+     * Returns a snapshot of the weather of the region owning the given location.
+     *
+     * @param loc the location whose chunk determines the region
+     * @return the current weather state of the region, or {@code null} if
+     * per-region weather is disabled for this world
+     * @throws IllegalStateException if not called from the region's tick thread
+     * @see #getRegionalWeather(World, int, int)
+     */
     default @Nullable RegionalWeatherSnapshot getRegionalWeather(Location loc) {
-        final int cx = loc.getBlockX() >> 4;
-        final int cz = loc.getBlockZ() >> 4;
-        return getRegionalWeather(loc.getWorld(), cx, cz);
+        final int chunkX = loc.getBlockX() >> 4;
+        final int chunkZ = loc.getBlockZ() >> 4;
+        return getRegionalWeather(loc.getWorld(), chunkX, chunkZ);
     }
 }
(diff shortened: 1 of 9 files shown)


canvas-server/src/main/java/io/canvasmc/canvas/world/weather/WeatherControllerImpl.java
```diff
@@ -0,0 +1,142 @@
+package io.canvasmc.canvas.world.weather;
+
+import com.google.common.base.Preconditions;
+import io.papermc.paper.threadedregions.RegionizedWorldData;
+import net.minecraft.server.level.ServerLevel;
+import org.bukkit.World;
+import org.bukkit.craftbukkit.CraftWorld;
+import org.jspecify.annotations.Nullable;
+
+public final class WeatherControllerImpl implements WeatherController {
```
Toffikk, line 16: Mark the class as @NullMarked and annotate each nullable type with @Nullable from jspecify

canvas-server/src/main/java/io/canvasmc/canvas/world/weather/WeatherControllerImpl.java
```diff
@@ -0,0 +1,142 @@
...
+        ca.spottedleaf.moonrise.common.util.TickThread.ensureTickThread(level, chunkX, chunkZ, "Cannot retrieve chunk asynchronously");
+
+        final RegionizedWorldData worldData = level.getCurrentWorldData();
+        final RegionizedWorldData.RegionalWeatherState st = worldData.weatherRegional;
+
+        if (st == null) {
+            return;
+        }
+
+        if (!st.initialized) {
```
Toffikk: try to use more descriptive names

canvas-api/src/main/java/io/canvasmc/canvas/world/weather/WeatherController.java
```diff
@@ -0,0 +1,24 @@
...
+import org.bukkit.Location;
+import org.bukkit.World;
+import org.jspecify.annotations.Nullable;
+
+public interface WeatherController {
+
+    void setRegionalWeather(World world, int chunkX, int chunkZ, WeatherType type, int durationTicks, boolean instantLevels);
+
+    default void setRegionalWeather(Location loc, WeatherType type, int durationTicks, boolean instantLevels) {
+        final int cx = loc.getBlockX() >> 4;
```
Toffikk: here too

canvas-server/src/main/java/io/canvasmc/canvas/world/weather/WeatherServer.java
```diff
@@ -0,0 +1,92 @@
+package io.canvasmc.canvas.world.weather;
+
+import io.papermc.paper.threadedregions.RegionizedWorldData;
+import net.minecraft.server.level.ServerLevel;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.util.RandomSource;
+
+public class WeatherServer {
```
Toffikk, line 16: @NullMarked

canvas-server/src/main/java/io/canvasmc/canvas/WorldConfig.java
```diff
@@ -678,4 +678,57 @@ public Component getSleepNotPossible() {
...
+                option("clear").docs("The weight for clear weather");
+                option("rain").docs("The weight for rain weather");
+                option("thunder").docs("The weight for thunder weather");
+            }
+            public int clear = 0;
+            public int rain = 1;
+            public int thunder = 2;
+        }
+
+        public MergeStrategy mergeStrategy = MergeStrategy.KEEP_INTO;
```
Toffikk, line 721: add docs for this, see above in the class for how it’s done

canvas-server/minecraft-patches/features/0003-Each-region-could-have-its-own-weather.patch
```diff
@@ -0,0 +1,504 @@
...
+     public void tick(BooleanSupplier haveTime, final io.papermc.paper.threadedregions.TickRegions.TickRegionData region) {
+         final io.papermc.paper.threadedregions.RegionizedWorldData regionizedWorldData = this.getCurrentWorldData();
+         regionizedWorldData.setHandlingTick(true);
+@@ -877,6 +878,17 @@ public class ServerLevel extends Level implements WorldGenLevel, ServerEntityGet
+             for (WorldBorder worldBorder : worldBorders) {
+                 worldBorder.tick();
+             }
++            // Canvas start - weather per-region
++            if (region.world.canvasConfig().weather.perRegion) {
++                final var st0 = regionizedWorldData.weatherRegional;
```
Toffikk: more descriptive names everywhere pls

Toffikk (changes requested):

Toffikk:
Generally appreciate this feature as it’s useful but the decision is up to @Dueris if we want to have such an invasive addition

Dueris:
I like the idea in concept, but a few notes:
- The plugin API to modify weather should set the weather for all regions
- The `weather` command should modify weather in all regions

Generally speaking, I like this idea. However I do question how exactly this would work with parity with Vanilla and the APIs. Would this work more as an override? How would this work to keep plugin compat? Im just moreso wanting to keep Vanilla compat and plugin API compat while doing something like this if that makes sense.

Idk, Im just more curious how this would work in terms of trying to make it so that it's compatible with plugins, while creating the system in question. It def could be useful for skyblock/oneblock servers.

Toffikk:
I think the weather command should affect the current region by default if executed by a player and the global one if by console. And we should also add a `global` argument so players can modify the weather globally

Toffikk:
Plugin API should work globally by default tho with more API added to be able to change it per region

Dueris:
Then would fetching the region(with no location specified) return the current region weather? And then setting it would set all regions? How would getting the region work from console or from a global perspective or the global tick?

Toffikk:
> Then would fetching the region(with no location specified) return the current region weather? And then setting it would set all regions? How would getting the region work from console or from a global perspective or the global tick?

It should fetch from the context, so if you’re on the global scheduler then global etc, same could be done for setting.

Dueris:
But if regions have their own weather, then what would it return from the global tick?

Toffikk:
> But if regions have their own weather, then what would it return from the global tick?

The last set global weather/default

Dueris:
What if it was never set? Would it return `CLEAR`?

Toffikk:
Yea

Toffikk:
I also think the per-region weather should be somewhat of an override, that all regions start out with the global weather whatever it happens to be and then change it on their own, and when new regions appear they should go through the same thing. So not eradicating global weather but treating it as the starting point

Dueris:
I agree

Dueris:
Should they fallback to the global weather too on occasion?

Toffikk:
> Should they fallback to the global weather too on occasion?

I don’t think so. It would be cleaner if on creation, the regional would snapshot the current global weather and then change independent of the global one, unless changed by the /weather global command which could set it for all regions

Dueris:
@Euphillya how does per-region weather persist across restarts?

Euphillya:
> @Euphillya how does per-region weather persist across restarts?

They aren't, given that the regions are dynamic. They are even reset to 0 ticks at that location:  https://github.com/Euphillya/Canvas/blob/2faf0cb26120391927365f40e25fa8b23de09a6d/canvas-server/minecraft-patches/features/0003-Each-region-could-have-its-own-weather.patch#L56

Dueris:
Yeahh I figured. Given this, it would technically deviate from Vanilla slightly with this, but I dont think this is a very important thing to be blunt. A restart resetting the weather doesnt seem that bad

BaconCat1:
When per region weather is off shouldn’t it behave exactly like vanilla and current canvas behavior though

Euphillya:
> When per region weather is off shouldn’t it behave exactly like vanilla and current canvas behavior though

Then, when the region is created, I can take the global world value and pass it on to the `WeatherServer`.



> When per region weather is off shouldn’t it behave exactly like vanilla and current canvas behavior though

If the configuration isn't enabled, it remains completely vanilla.

canvas-server/src/main/java/io/canvasmc/canvas/WorldConfig.java
Euphillya: I added it here—is it in the wrong place? Image

BaconCat1:
> > When per region weather is off shouldn’t it behave exactly like vanilla and current canvas behavior though
> 
> 
> 
> Then, when the region is created, I can take the global world value and pass it on to the `WeatherServer`.
> 
> 
> 
> 
> 
> 
> 
> > When per region weather is off shouldn’t it behave exactly like vanilla and current canvas behavior though
> 
> 
> 
> If the configuration isn't enabled, it remains completely vanilla.

Good idea!

Toffikk:
genuinely worst news of today😓

Euphillya:
> genuinely worst news of today😓

It's mainly because I haven't touched the PR in quite a while, and rather than leaving it open, I'd prefer to close it.

PR #288, opened 2026-07-11
Reorganize patches

Biquaternions:
This PR makes the following changes:
1. Drops async-protocol-switch
It is a bad approach, it introduces a race condition where players might not receive some packets while the output channels are being configured, as it's now made async.
2. Refactors the region threading base patch, isolating https://github.com/CraftCanvasMC/Canvas/pull/164 into its own feature patch, and including proper credits.
3. Restores William as the original author of "Purpur Ender Chest 6 Rows Config".


BaconCat1:
But at that point imo we should only credit the small part that is theirs, because they don't own the idea. And much of the code surrounding it is my own and Dueris unique idea on how it should work.

BaconCat1:
Along with fixing the issues of it

Dueris:
Id like to keep the cross region single tick movement in the region threading patch. Add a comment explaining the source in the parts where the code is patched. Beyond that this is fine

Biquaternions:
The cross region single tick is still there, I just moved it into 0008-Fix-CME... to add the corresponding patch notes and attributions

Dueris:
While I understand that, I was moreso meaning to add those notes and attributions within the source and at the head of the region threading patch. May also be nice to clarify in the region threading patch it is a combination of Canvas fixes, Folia's base, and Luminol fixes. It was on my TODO for when I was gonna do another repo clean(like the README needed work) but Ig no time like the present...

Dueris:
Id like to keep the CME fix in the region threading patch primarily because it keeps it all together

Biquaternions:
Merged back to the region threading base patch.
I added the attribution notice on the conflicting code, although it's literally just a teleport, there's genuinely no way for those to not be similar.

PR #289, opened 2026-07-11
Fix portal momentum

LuckyODO:
Fixes entity velocity being lost during Canvas's asynchronous portal teleportation.

This issue caused:

- Sand dupers to send sand to the End without the expected outward spread
- Items and primed TNT to lose momentum after using Nether portals
- TNT to stop at the portal exit and explode in place
- Some projectiles to continue ticking at their old position after teleporting


R00tB33rMan:
We (the CanvasMC team) are closing this PR due to the improper attribution and credit from the original source of this code, LuminolMC, and improper patching of this as well.

These changes should be applied in the `0004-Region-Threading` patch, alongside that multiple of your comments regarding where Canvas makes changes are wrongly placed and also not annotated at all in some places, alongside deliberately removing Luminol as the source author of the changes in question presented by this PR. Your PR title also is misleading, given it does not include the entirety of what this PR offers.

PR #292, opened 2026-07-15
Remove ineffective chunkgen opts 

BaconCat1:
This PR removes the following chunkgen optimizations which are generally ineffective and in some cases detrimental on modern versions.
Screenshot 2026-07-14 at 5 53 35 PM
As shown in this testing graph, at 4 chunk workers they tend to be neutral or detrimental, and at 10 workers some of them provide marginal benefit.

I have also tested them at 16 and 20 workers and the results were as follows:
Screenshot 2026-07-14 at 6 20 07 PM

As shown, even when scaling to larger thread counts, the benefits of these are pretty marginal.

PR #293, opened 2026-07-15
Enforce documentation for configs via test suites

Toffikk:
This pull request adds tests for the global and world configuration classes which enforce all options have documentation attached to them.
The verification can be opted out of by annotating the relevant config with `@Undocumented` along with a reason.

In addition to above it completes https://panel.canvasmc.io/dashboard/tasks/cmr0gyjxp00ap0klpfncznk15
___
#### Example output
image
image


Dueris:
I'm not sure about this. The configs that don't have a config are sorta supposed to be the ones that are self explanatory as to avoid cluttering the config with a ton of documentation(bc then it looks too busy)

Toffikk:
> I'm not sure about this. The configs that don't have a config are sorta supposed to be the ones that are self explanatory as to avoid cluttering the config with a ton of documentation(bc then it looks too busy)

Then they can just be annotated as Undocumented like in this pr, perhaps with a more fitting reason or a simple `doesn’t require docs`

Toffikk:
This is more so as a safeguard to make sure config options that don’t have comments are intended not to have them. With this change i discovered some like purpur alt keepalive not being documented

Toffikk:
Blocked by: #292

canvas-server/src/main/java/io/canvasmc/canvas/GlobalConfiguration.java
```diff
@@ -551,6 +553,13 @@ public static class Networking extends Part {
                     "If alternative playerlist tick is enabled, this is the interval in ticks for how often",
                     "each bucket will be ticked"
                 ).greaterThan(0.0F);
+            option("purpurAlternativeKeepalive")
+                .docs(
```
Dueris, line 519: This method(`docs`) wraps the text in the blob. Probably best to use a `Style` instance and make the `AKA` specifically on a new line rather than wrapped with everything else.

canvas-server/src/main/java/io/canvasmc/canvas/WorldConfig.java
```diff
@@ -486,14 +496,19 @@ public static class Mace extends Part {
             option("useLegacyBlastProtection").docs("Restores the blast protection logic from before 1.21");
         }
 
+        @Undocumented("Doesn't require docs.")
         public boolean disableSweepingEdge = false;
+        @Undocumented("Doesn't require docs.")
         public boolean disableCritsWhileSprinting = false;
         // public int invulnerabilityTicks = 10;
```
Dueris: Perhaps out of scope of this PR but we can probably remove this comment too since there isn't a reliable way to reintroduce this given our new setup

canvas-server/src/test/java/io/canvasmc/canvas/configuration/GlobalConfigurationDocumentationTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+        for (final Field field : partClass.getDeclaredFields()) {
+            if (!Modifier.isPublic(field.getModifiers())
+                || Modifier.isStatic(field.getModifiers())
+                || field.isSynthetic()) {
+                continue;
+            }
+
+            if (Part.class.isAssignableFrom(field.getType())) {
+                @SuppressWarnings("unchecked")
```
Dueris: Use a comment to suppress the warnings to follow the rest of the code-style:
```
//noinspection unchecked
```

canvas-server/src/test/java/io/canvasmc/canvas/configuration/VanillaFixesConfigurationTest.java
```diff
@@ -0,0 +1,199 @@
...
+        if (issues.isEmpty()) {
+            return;
+        }
+
+        final JsonArray found = queryMojira(issues);
+
+        final Set<String> returned = new HashSet<>();
+        final Set<String> fixed = new HashSet<>();
+
+        for (final var element : found) {
```
Dueris: While I honestly dont care TOO much, use the actual type instead of `var` lol

Dueris (dismissed):
After these are addressed and Bacons other PR that is currently blocking this is merged, Im good with this being merged too

Issue #303, opened 2026-07-28
NPE player crash with unitialized score on player join on the server after player changes dimensions (?)


DenisMasterHerobrine:
### Stack trace

First reported evidence stack-trace to me.

```
net.minecraft.ReportedException: Ticking player
	at net.minecraft.server.level.ServerPlayer.doTick(ServerPlayer.java:1138) ~[canvas-26.2.jar:26.2-878-28c39ae]
	at net.minecraft.server.network.ServerGamePacketListenerImpl.tickPlayer(ServerGamePacketListenerImpl.java:400) ~[canvas-26.2.jar:26.2-878-28c39ae]
	at net.minecraft.server.network.ServerGamePacketListenerImpl.tick(ServerGamePacketListenerImpl.java:379) ~[canvas-26.2.jar:26.2-878-28c39ae]
	at net.minecraft.network.Connection.tick(Connection.java:641) ~[canvas-26.2.jar:26.2-878-28c39ae]
	at io.papermc.paper.threadedregions.RegionizedWorldData.tickConnections(RegionizedWorldData.java:549) ~[canvas-26.2.jar:26.2-878-28c39ae]
	at net.minecraft.server.MinecraftServer.tickChildren(MinecraftServer.java:1889) ~[canvas-26.2.jar:26.2-878-28c39ae]
	at net.minecraft.server.MinecraftServer.tickServer(MinecraftServer.java:1669) ~[canvas-26.2.jar:26.2-878-28c39ae]
	at net.minecraft.server.dedicated.DedicatedServer.tickServer(DedicatedServer.java:420) ~[canvas-26.2.jar:26.2-878-28c39ae]
	at io.papermc.paper.threadedregions.TickRegions$ConcreteRegionTickHandle.tickRegion(TickRegions.java:615) ~[canvas-26.2.jar:26.2-878-28c39ae]
	at io.papermc.paper.threadedregions.TickRegionScheduler$RegionScheduleHandle.runTick(TickRegionScheduler.java:604) ~[canvas-26.2.jar:26.2-878-28c39ae]
	at ca.spottedleaf.concurrentutil.scheduler.EDFSchedulerThreadPool$TickThreadRunner.run(EDFSchedulerThreadPool.java:528) ~[leafpile-1.0.0.jar:26.2-878-28c39ae]
	at ca.spottedleaf.concurrentutil.util.LazyRunnable.run(LazyRunnable.java:25) ~[leafpile-1.0.0.jar:?]
[...]

PR #304, opened 2026-07-29
Convert issue template preflight prompts to checkboxes

BaconCat1:
The checkboxes were just markdown and you literally couldn't click it 😭.

This PR fixes that. 

Testing:

Trust me bro.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -5,7 +5,7 @@ contact_links:
     about: Ask setup, configuration, and general support questions on Discord.
   - name: Security vulnerability
     url: https://canvasmc.io/discord
-    about: Report severe vulnerabilities privately to someone with the blue @Development Team role at the top of the member list; do not disclose exploit details publicly.
+    about: For severe vulnerabilities only, privately contact someone with the blue @Development Team role near the top of the member list. Do not disclose exploits publicly; otherwise, please don't ping us.
   - name: Documentation
     url: https://docs.canvasmc.io/
     about: Review the Canvas documentation before opening an issue.
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,11 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Canvas support
+    url: https://canvasmc.io/discord
+    about: Ask setup, configuration, and general support questions on Discord.
+  - name: Security vulnerability
+    url: https://discord.gg/uM7eC7rBrW
+    about: Report severe vulnerabilities privately—do not disclose exploit details publicly.
+  - name: Documentation
+    url: https://docs.canvasmc.io/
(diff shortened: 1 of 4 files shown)


Dueris (approved):

BaconCat1:
Note before merging to review [`08cc940c`](https://github.com/CraftCanvasMC/Canvas/commit/08cc940c749e8d95cc93edb03f96a6b9d9b839e4), which prevents bypassing issue templates.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,11 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Canvas support
+    url: https://canvasmc.io/discord
+    about: Ask setup, configuration, and general support questions on Discord.
+  - name: Security vulnerability
+    url: https://discord.gg/uM7eC7rBrW
```
Dueris: This should also be `https://canvasmc.io/discord`, as we don't want this to expire for whatever reason(even if it's not already going to expire, we should set it to the link we already use everywhere which is what `https://canvasmc.io/discord redirects you to)

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,11 @@
...
+    about: Report severe vulnerabilities privately—do not disclose exploit details publicly.
```
Dueris: Clarify that you should ask one of the people with the `@Development Team` role(the blue one at the top above everyone else)

Dueris (changes requested):

Dueris (approved):

R00tB33rMan (approved):

Dueris:
We should change our issue report templates probably to include the datapack command if not already included. The #303 issue report made me think of adding it back since datapacks can be useful in finding the cause of issues and stuff, and that issue was *sorta* induced by a datapack but also because of an issue with Canvas' scoreboard rewrite. The only caveat is that the datapack command in Canvas now is only able to read the enabled packs, since scanning for new ones or even enabling them at runtime is like... very big nono in Folia. Anyway, if it's not already included, perhaps that could be added here in this PR

BaconCat1:
> We should change our issue report templates probably to include the datapack command if not already included. The #303 issue report made me think of adding it back since datapacks can be useful in finding the cause of issues and stuff, and that issue was *sorta* induced by a datapack but also because of an issue with Canvas' scoreboard rewrite. The only caveat is that the datapack command in Canvas now is only able to read the enabled packs, since scanning for new ones or even enabling them at runtime is like... very big nono in Folia. Anyway, if it's not already included, perhaps that could be added here in this PR

I'm all in support of this but I'm not 100% sure I understand what exactly I should add lol.

Dueris (approved):
We already include the `/datapack list` info request in our issue template, which is included here, so you can ignore my previous comment. No clue why this was in here already, prob bc this was based on Papers issue template initially, but it doesn't entirely matter as long as it's in there.

PR #305, opened 2026-07-29
Add github actions checks for pull requests

BaconCat1:
Adds github actions workflows to validate PRs.

- Applying all Canvas patches
- Compiling it
- Running the test suite (`./gradlew test`)


Dueris:
I think one thing we should verify before we merge this is that our tests actually pass, since I genuinely dont know if we pass Papers tests, since we dont currently use them in our CI(which we are planning to change)

Dueris:
Also, why did you add multiple workflows for compiling and applying patches? Wouldnt the `test` workflow cover both of those extra ones? Just seems unnecessary

BaconCat1:
> Also, why did you add multiple workflows for compiling and applying patches? Wouldnt the `test` workflow cover both of those extra ones? Just seems unnecessary

So that you can see at a glance what actually failed and so can the pr author, instead of having to actually go investigate the action run itself.

Dueris:
I feel like maybe having 1 action run instead of like 3 being run is maybe a bit better than having 3 run every single time IMHO

BaconCat1:
> I feel like maybe having 1 action run instead of like 3 being run is maybe a bit better than having 3 run every single time IMHO

Alright it's gone

.github/workflows/test.yml
```diff
@@ -0,0 +1,38 @@
+name: Test
```
Dueris: Maybe a better name like "Test PR" or something rather than just... "Test"

Dueris (changes requested):
Tiny last nitpick, after that LGTM

.github/workflows/test.yml
BaconCat1: Renamed the workflow - can't rename the file from my phone.

BaconCat1:
Done - I renamed the file too.

Dueris (approved):

Toffikk (approved):

PR #310, opened 2026-08-01
Introduce dual-queue backed queue for the Affinity scheduler

Dueris:
Original concept idea created by Jan.

This concept essentially introduces a 2nd priority queue ordered by the last insertion time, which is used later on to process intermediate tasks if the current tick we are waiting for does not have tasks to process. This improves efficiency of the scheduler and intermediate tasks, reducing latency of packet processing, chunk tasks, and tick tasks in the region.

The new queue implementation is the `SchedulableTickFastQueue`, which is backed by 2 `FastHeapPriorityQueue` instances, `edf` and `queueTime`. `edf` is ordered by EDF(earliest deadline first), maintaining the queue order for what tick needs to be processed next. `queueTime` is ordered by the insertion time of the schedulable tick into the EDF queue, with the head always being the task that was inserted the longest time ago.

This should prove to be very useful and powerful for a lot of servers. Further optimizations will be done down the line to try and maximize the time spent processing intermediate tasks through load balancing optimizations that will be done at a later date.

This marks [this panel task](https://panel.canvasmc.io/dashboard/tasks/cms7kk0zg002q0jlplycrw4ad) as in-progress


R00tB33rMan (approved):

Toffikk (approved):

devnootnoot:
yep, loved the original idea by Jan, looks like it's implemented nicely

devnootnoot (approved):

PR #315, opened 2026-08-07
feat: parallel TNT explosion ray tracing for Folia multi-core

Cheerimy-Studio:
## Summary

Adds parallel TNT explosion ray tracing for Folia/Canvas multi-core servers.

### What it does
- Parallelizes calculateExplodedPositions() (1352 rays per TNT explosion) across multiple CPU cores using ForkJoinPool work-stealing
- Only affects PrimedTnt explosions — creepers, beds, crystals, ender dragons, etc. remain unchanged
- Default **enabled** via optimized-explosions.enabled: true in canvas-server.yml

### What it does NOT change
- hurtEntities() — vanilla CraftBukkit entity damage + EntityExplodeEvent unchanged
- interactWithBlocks() — block destruction, drops, chain reactions unchanged
- createFire() — fire spawning unchanged

### Technical details
- **Thread safety**: Each parallel batch uses its own block cache (no shared mutable state)
- **ThreadLocalRandom**: Replaces level.getRandom() (not thread-safe) for ray power randomization
- **Chunk pre-loading**: All chunks in explosion radius are pre-loaded on the region thread to prevent async chunk loading
- **Smart thread pool**: min(cores, max(4, cores*3/4)) — avoids starving Folia/Moonrise/GC
- **5-second timeout per batch**: Single batch failure doesn't block the explosion

### Files changed
| File | Change |
|------|--------|
| canvas-server/.../AsyncExplosionEngine.java | Parallel ray tracing engine (new, rewritten) |
| canvas-server/.../ServerExplosion.java.patch | Injection point in explode() method |
| canvas-server/.../GlobalConfiguration.java | optimized-explosions config section |

### Testing
Build with ./gradlew :canvas-server:createBundlerJar and test with TNT in a multi-core server.  
Config toggle: optimized-explosions.enabled: false to disable.


BaconCat1:
The comments will need to be translated into English.

Dueris:
It's also worth noting that the level random source is threadsafe as it was replaced with a `ThreadLocalRandomSource` instance provided by Folia.

Im unsure on this PR in general, given Folia sorta isn't the best place for making async operations given just the architecture of region threading, and making more things async can actually be a performance *problem*, not an optimization.

Note for others reviewing, the PR was created with an AI but code was created by the human from my understanding. Wanted to clarify that so none of the other maintainers close this up

Cheerimy-Studio:
It work on E5 well.

Cheerimy-Studio:
I only can wrote plugins, this patch made by AI, it is have some problems please tell me. Thank.

canvas-server/minecraft-patches/features/0003-Async-TNT-Explosion.patch
```diff
@@ -0,0 +1,45 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Cheerimy <cheerimy@canvasmc.io>
```
BaconCat1, line 2: Not sure how I feel about this one tbh

canvas-server/minecraft-patches/features/0003-Async-TNT-Explosion.patch
```diff
@@ -0,0 +1,45 @@
...
+ package net.minecraft.world.level;
++// Canvas start - async tnt explosion
+ 
+ import com.google.common.collect.ImmutableList;
+ import com.google.common.collect.Lists;
+@@ -19,6 +20,7 @@ import net.minecraft.world.entity.EntityTypes;
+ import net.minecraft.world.entity.LivingEntity;
+ import net.minecraft.world.entity.Marker;
+ import net.minecraft.world.entity.ai.attributes.Attributes;
++import net.minecraft.world.entity.item.PrimedTnt;
```
BaconCat1, line 21: We also shouldn't use imports here; use fully qualified name where this is used.

canvas-server/minecraft-patches/features/0003-Async-TNT-Explosion.patch
BaconCat1: Make it sound like he is on core canvas team.

Dueris: The other thing is, we dont even use that address for emails at all. It was a discussion internally but we dont use that address at all for emails

BaconCat1:
I’m also unsure about my feelings on this. The purpose of folia/canvas is to prevent any single region from causing the entire server to slow down or lag. However, if a region experiences a high amount of TNT activity, it can use significantly more resources with this pr, potentially causing other regions to suffer as well.

canvas-server/minecraft-patches/features/0003-Async-TNT-Explosion.patch
Cheerimy-Studio: I am sorry, that made by Agent.
I only know how to make plugins

BaconCat1: > I am sorry, that made by Agent.
> 
> I only know how to make plugins

...... not sure how setting the email is hard but ok....

Cheerimy-Studio: I am a Chinese and in China we need to use VPN to connect Github.
So i did not know how to use git, that is Git by Agent.

Cheerimy-Studio: In my server on 44H server, with 10000 tnt will use 8 core of server.

BaconCat1: > I am a Chinese and in China we need to use VPN to connect Github.
> 
> So i did not know how to use git, that is Git by Agent.

All the Chinese GitHub equivalents still use git tho

canvas-server/minecraft-patches/features/0003-Async-TNT-Explosion.patch
Cheerimy-Studio: How can i change it?

BaconCat1: Instead of import use fully qualified name - for example

// instead of:
import example.entity.Creeper;

Creeper creeper = new Creeper();

// Use:
example.entity.Creeper creeper = new example.entity.Creeper();

Dueris:
There are many things to comment on in this PR. TLDR, we are closing this, we do not think this is a good fit for Canvas or Folia servers. To explain more on our reasoning:

- The current PR has no patch for this, just the newly added class, config changes, and a random `applyPatches.log` file, which should not be in this PR.
- In your previous commit, before deleting your patch wholesale, you utilized the email of `cheerimy@canvasmc.io`, which is not only not a real GitHub account, but also an invalid email as we don't even use this domain for our own emails, and you are not part of the Canvas team to be allowed to utilize an email with that domain.
- This seems largely AI-generated programming. This is not something we really like, as we expect the creator of the PR to actually have reviewed the code in full and monitored over the AI or used it as an assistive tool, rather than a replacement wholesale for development work.
- Folia/Canvas are not really the best places for async work. Part of the reason why we don't make parts of the tick async is because this can actually hurt performance with Folia-based software rather than help it. So this wouldn't be a great addition to Folia/Canvas anyways
- In response to https://github.com/CraftCanvasMC/Canvas/pull/315#discussion_r3733835281, we expect you, as the author of the PR, to have some understanding of the patch system to create and make changes to your PR. This is like.. required of sorts.

In general, we feel it is best this not be merged into Canvas. We encourage you to still PR here in the future, as we would love to have you as a contributor, but I think this PR is not something that would be good for Canvas. Thank you

Issue #318, opened 2026-08-09
Worldgen optimizations


personalized-advertising:
### Before submitting

- [x] I'm using a [supported version of Canvas](https://docs.canvasmc.io/canvas/lts/).
- [x] The feature I'm requesting isn’t already included in the version I'm using.
- [x] I'm clicking all these boxes without reading them
- [x] I'm an AI agent filling out this feature request
- [x] I have searched for and confirmed there isn’t already [an open request](https://github.com/CraftCanvasMC/Canvas/issues?q=is%3Aissue%20is%3Aopen%20type%3AFeature) for this feature.

### Is your feature request related to a problem?

World generation is the biggest bottleneck in terms of CPU, a few mods suc has c2me have world gen optimizations:

https://gist.github.com/andreypfau/e490c562ec0ea3d61d58bd9799fe1463#1-density-function-compiler-c2me-opts-dfc

### Describe the solution you'd like.

Implement some optimizations from c2me for world gen such as the density function compiler.

### Describe alternatives you've considered.

buying better CPU

### Issue search

- [x] I have searched the issue tracker for similar or identical requests.

Dueris:
It appears as if you just... clicked through everything... nice...

The DFC is not something Canvas is looking to implement anytime soon. The DFC has proven to be largely ineffective in Paper-based software and at times can even slow down CPS from testing Ive done. Then again, both C2ME and Moonrise are completely different architectures, so it probably benefits them more than it would Paper/Moonrise.

Regardless, the DFC is not something we want to add.

Also, please actually read the boxes you are checking, not blindly checking them...

personalized-advertising:
> It appears as if you just... clicked through everything... nice...
> 
> The DFC is not something Canvas is looking to implement anytime soon. The DFC has proven to be largely ineffective in Paper-based software and at times can even slow down CPS from testing Ive done. Then again, both C2ME and Moonrise are completely different architectures, so it probably benefits them more than it would Paper/Moonrise.
> 
> Regardless, the DFC is not something we want to add.
> 
> Also, please actually read the boxes you are checking, not blindly checking them...

Mb the boxes are always BS so I never read them, plus its a feature request lol

personalized-advertising:
> It appears as if you just... clicked through everything... nice...
> 
[...]

PR #334, opened 2026-09-09
Port Leaf’s C2ME density-function compiler and worldgen micro-opts

ssquadteam:
- Port Leaf’s C2ME DensityFunction compiler. It JIT-compiles noise-router trees to bytecode when a dimension first builds its router. Off by default: `chunk-system.density-function-compiler` in `config/canvas-server.yml`. Enabling after worlds are already loaded does not compile those dimensions; the flag must be true at world load.
- Port Leaf’s C2ME ore BitSet pool (`OreFeature`) so vein placement reuses per-thread BitSets instead of allocating a new one per blob.
- Port Leaf’s C2ME Beardifier / structure-weight math: array iteration, one switch per piece, and a cheaper bury falloff.
- Cache world-generator sea level and biome zoom seed so worldgen stops repeating those lookups.
- Finish the leftover C2ME noise / getBiome bits Canvas did not already have (PerlinNoise.wrap, noiseWithDerivative, quart-offset table).
- Skip the End surface pass when the rule is already a single default End-stone block. Config: chunk-system.optimize-end-surface-generation (on by default).

### DIFF
--- a/canvas-server/minecraft-patches/sources/net/minecraft/server/MinecraftServer.java.patch
+++ b/canvas-server/minecraft-patches/sources/net/minecraft/server/MinecraftServer.java.patch
@@ -1,5 +1,15 @@
 --- a/net/minecraft/server/MinecraftServer.java
 +++ b/net/minecraft/server/MinecraftServer.java
+@@ -680,8 +_,7 @@
+         final ResourceKey<Level> dimensionKey = loading.info().dimensionKey();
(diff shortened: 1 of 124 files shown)


personalized-advertising:
lgtm

Biquaternions:
Mojang already included a density function compiler in 26.3
Also, the idea of porting C2ME's implementation has already been rejected before.
See: https://github.com/CraftCanvasMC/Canvas/issues/318

Dueris:
@Biquaternions is correct.

PR #335, opened 2026-09-12
Extended View Distance

Dueris:
This will be made into a base patch soon, but I need to extract the per-world-distance system too, since this is based somewhat on that system.

This adds a new "fake chunk" system to Canvas, sending out fake chunks to extend the view distance of the client without actually loading any chunks on the server. One caveat is that this *only* reads already generated chunks from disk, and does *not* generate new chunks, meaning this will only really be beneficial for servers that are pre-generated.

This needs some production testing, and a few more things need to be done, but the base implementation is done. This also has builtin "hollowing" of chunks to try and set as much of the chunk to air as possible to try and reduce bandwidth as much as possible. At the time of creating this PR bandwidth for these chunks seems to be cut anywhere between 35-65%, and hopefully it will be a bit better later.

### DIFF
--- a/canvas-server/minecraft-patches/features/0003-Extended-View-Distance.patch
+++ b/canvas-server/minecraft-patches/features/0003-Extended-View-Distance.patch
@@ -1,0 +1,1209 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Dueris <duerismc@gmail.com>
+Date: Fri, 11 Sep 2026 15:52:13 -0700
+Subject: [PATCH] Extended View Distance
+
+
+diff --git a/ca/spottedleaf/moonrise/patches/chunk_system/player/RegionizedPlayerChunkLoader.java b/ca/spottedleaf/moonrise/patches/chunk_system/player/RegionizedPlayerChunkLoader.java
+index fad788a6541a7e9edd42f699d2968897d446a353..ab35b21829601da927b6b882c674cf6659fb4254 100644
+--- a/ca/spottedleaf/moonrise/patches/chunk_system/player/RegionizedPlayerChunkLoader.java
++++ b/ca/spottedleaf/moonrise/patches/chunk_system/player/RegionizedPlayerChunkLoader.java
+@@ -63,7 +63,7 @@ public final class RegionizedPlayerChunkLoader {
+         private static final VarHandle VIEW_DISTANCES_HANDLE = ConcurrentUtil.getVarHandle(ViewDistanceHolder.class, "viewDistances", ViewDistances.class);
+ 
+         public ViewDistanceHolder() {
+-            VIEW_DISTANCES_HANDLE.setVolatile(this, new ViewDistances(-1, -1, -1));
++            VIEW_DISTANCES_HANDLE.setVolatile(this, new ViewDistances(-1, -1, -1, -1)); // Canvas - extended view distance
+         }
+ 
+         public ViewDistances getViewDistances() {
+@@ -105,6 +105,14 @@ public final class RegionizedPlayerChunkLoader {
+                 return param.setSendViewDistance(distance);
+             });
+         }
++        // Canvas start - extended view distance
++
++        public void setVisualViewDistance(final int distance) {
++            this.updateViewDistance((final ViewDistances param) -> {
++                return param.setVisualViewDistance(distance);
++            });
++        }
++        // Canvas end - extended view distance
+ 
+         public JsonObject toJson() {
+             return this.getViewDistances().toJson();
+@@ -114,13 +122,16 @@ public final class RegionizedPlayerChunkLoader {
+     public static final record ViewDistances(
+         int tickViewDistance,
+         int loadViewDistance,
+-        int sendViewDistance
++        // Canvas start - extended view distance
++        int sendViewDistance,
++        int vvDistance
++        // Canvas end - extended view distance
+     ) {
+         public ViewDistances setTickViewDistance(final int distance) {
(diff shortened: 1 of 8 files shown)


canvas-server/minecraft-patches/features/0003-Extended-View-Distance.patch
```diff
@@ -0,0 +1,1209 @@
...
++
++    public static int getAPIVisualViewDistance(final ServerPlayer player) {
++        final ServerLevel level = player.level();
++        final PlayerChunkLoaderData data = ((ChunkSystemServerPlayer)player).moonrise$getChunkLoader();
++        if (data == null) {
++            return ((ChunkSystemServerLevel)level).moonrise$getPlayerChunkLoader().getAPIVisualViewDistance();
++        }
++        return data.lastVVDistance;
++    }
++    // Canavs end - extended view distance
```
R00tB33rMan: Found typo

R00tB33rMan (changes requested):

canvas-server/minecraft-patches/features/0003-Extended-View-Distance.patch
```diff
@@ -0,0 +1,1209 @@
...
++import org.jspecify.annotations.Nullable;
++
++public class PacketConstructorUtils {
++
++    public static final int MASK_SIZE = 4096; // 16^3
++
++    private static final Predicate<BlockState> IS_ORE = (state) ->
++        state.is(BlockTags.COPPER_ORES)
++        || state.is(BlockTags.IRON_ORES)
++        || state.is(BlockTags.GOLD_ORES)
```
R00tB33rMan, line 775: A few notes that we'll need to likely tackle:
OLD_ORES (which includes nether gold ore) is replaced with STONE in the nether while ancient debris gets the netherrack special case, also that emerald ore, deepslate emerald ore and nether quartz ore are not in IS_ORE, given fake packets skip anti-xray entirely, and that the "bottom half is truncated" assumption in clearOres only holds when hollowChunks is on, so hideOres alone puts stone specks into deepslate

canvas-server/minecraft-patches/features/0003-Extended-View-Distance.patch
```diff
@@ -0,0 +1,1209 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Dueris <duerismc@gmail.com>
+Date: Fri, 11 Sep 2026 15:52:13 -0700
+Subject: [PATCH] Extended View Distance
+
+
+diff --git a/ca/spottedleaf/moonrise/patches/chunk_system/player/RegionizedPlayerChunkLoader.java b/ca/spottedleaf/moonrise/patches/chunk_system/player/RegionizedPlayerChunkLoader.java
+index fad788a6541a7e9edd42f699d2968897d446a353..fede09ee32da72d414dacc553af91199e954ad99 100644
+--- a/ca/spottedleaf/moonrise/patches/chunk_system/player/RegionizedPlayerChunkLoader.java
++++ b/ca/spottedleaf/moonrise/patches/chunk_system/player/RegionizedPlayerChunkLoader.java
```
R00tB33rMan, line 10: Note: We intentionally do not latch onto the `PlayerChunkLoadEvent`/`PlayerChunkUnloadEvent`, since they do not relevantly apply to the server and shouldn't be called by plugins.

R00tB33rMan (commented):
Raw region NBT is parsed without `chunkMap.upgradeChunkTag` first, unlike `ChunkLoadTask.runOffMain`, so chunks saved under an older data version render wrong or throw. (Currently uncertain)

canvas-server/minecraft-patches/features/0003-Extended-View-Distance.patch
Dueris: Intentional, I do not want any API touching this.

canvas-server/minecraft-patches/features/0003-Extended-View-Distance.patch
Dueris: This is something I want to improve in general via a customizable map of sorts in the config to let servers configure their own block "hide" or "replace" mechanics.

Dueris:
> Raw region NBT is parsed without chunkMap.upgradeChunkTag first, unlike ChunkLoadTask.runOffMain, so chunks saved under an older data version render wrong or throw. (Currently uncertain)

You are correct. Fixed.

Cubicake:
Will there be a way to detect fake chunks from the plugin-side? I do a lot of chunk packet modifying in my plugin and I think it'd be better to skip that logic for fake chunks given that the more authoritative chunk packet will be sent if the player gets close.

Dueris:
You could check the `VVChunkCache` of the world to see if the fake chunk is present in the cache. I plan to do a lot more "hollowing" and generic culling of the chunk in general though. What sort of thing r u needing for your plugin

Cubicake:
> What sort of thing r u needing for your plugin

Just some sort of lightweight way to flag it as a fake chunk so I can bypass my expensive logic for it. Preferably it'd be one of
1. Data attached to the chunk
  a. Admittedly this would be hacky
2. Some sort of player-netty-thread local data store I can query with chunk coordinates and receive whether it is a real, fake, or unloaded chunk for that player
3. Some sort of event that fires on the player's netty thread so I can make my own version of option 2

Dueris:
Do u have access to the NMS object instance of the packet in your plugin? `ClientboundLevelChunkWithLightPacket`

Cubicake:
I'm using packetevents so unfortunately not currently. If you decide to store the flag on the NMS object though, I could add a canvas-specific path to access it and check for the flag then implement option 2 myself, so if that's the easiest way then it works for me.

